PR #2, opened 2016-07-09
Add rpmsg_piksi package

jacobmcnamee:
Device driver used to map firmware RPMSG endpoints to character devices.


jacobmcnamee:
TODO: Maintain packet boundaries

PR #6, opened 2016-07-25
Speed up build process

jacobmcnamee:
- Switch to downloaded external toolchain (saves building the compiler)
- Replace ping loop with `travis_wait`


jacobmcnamee:
:+1: `Elapsed time 18 min 30 sec`

PR #10, opened 2016-08-09
Forward signals to child processes

jacobmcnamee:
This should cause all the forked children to close when a signal is sent to the parent.


jacobmcnamee:
Now with 100% less killing other unrelated processes...

PR #15, opened 2016-08-16
Upload images from travis build to S3

margaret:
Part of https://github.com/swift-nav/test_team_planning/issues/107
/cc @cbeighley @jacobmcnamee

### DIFF
--- a/upload_artifacts.sh
+++ b/upload_artifacts.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Used in .travis.yml
+
+BUCKET=piksi-buildroot-images
+PIKSI_VERSION=v3
+
+set -e
+
+# Install the AWS command line interface
+curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
+unzip awscli-bundle.zip
+sudo ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
+
+BUILD_VERSION=$(git rev-parse --short HEAD)
+BUILD_DIR=commit_$BUILD_VERSION
+mkdir -p $BUILD_DIR
+
+# "folders" on S3 are prefix keys
+aws s3api put-object --bucket $BUCKET --key $PIKSI_VERSION/$BUILD_DIR/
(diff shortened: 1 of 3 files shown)


upload_artifacts.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Used in .travis.yml
+
+BUCKET=piksi-buildroot-images
+PIKSI_VERSION=v3
+
+set -e
+
+# Install the AWS command line interface
+curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
```
mfine: Instead, should be able to use:

```
before_install:
  - pip install --user --upgrade awscli
```

upload_artifacts.sh
```diff
@@ -0,0 +1,29 @@
...
+unzip awscli-bundle.zip
+sudo ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
+
+BUILD_VERSION=$(git rev-parse --short HEAD)
+BUILD_DIR=commit_$BUILD_VERSION
+mkdir -p $BUILD_DIR
+
+# "folders" on S3 are prefix keys
+aws s3api put-object --bucket $BUCKET --key $PIKSI_VERSION/$BUILD_DIR/
```
mfine: What is this doing?

upload_artifacts.sh
```diff
@@ -0,0 +1,29 @@
...
+UPLOAD_DIR="$BUCKET/$PIKSI_VERSION/$BUILD_DIR"
+
+cp "./output/images/boot.bin" "$BUILD_DIR"
```
mfine: I'd just make 4 `aws s3 cp` calls vs. cp things into a folder, and then copying the folder to s3.

upload_artifacts.sh
margaret: It's creating the "folder" for the build. I tried a whole bunch of different ways but that was the only one I could get to work. http://stackoverflow.com/questions/33113354/how-to-create-a-folder-like-i-e-prefix-object-on-s3-using-the-aws-cli

PR #16, opened 2016-08-18
Update travis upload to work with anonymous-writable bucket

margaret:
- pull requests now also upload images to S3

### DIFF
--- a/upload_artifacts.sh
+++ b/upload_artifacts.sh
@@ -1,21 +1,33 @@
 #!/bin/bash
 
-# Used in .travis.yml
-
-BUCKET=piksi-buildroot-images
-PIKSI_VERSION=v3
+# This script is used in .travis.yml
 
 set -e
 
+BUCKET=piksi-buildroot-images
+PIKSI_VERSION=v3
 BUILD_VERSION=$(git rev-parse --short HEAD)
-BUILD_DIR=commit_$BUILD_VERSION
-# "folders" on S3 are prefix keys
-# http://stackoverflow.com/questions/33113354/how-to-create-a-folder-like-i-e-prefix-object-on-s3-using-the-aws-cli
-aws s3api put-object --bucket $BUCKET --key $PIKSI_VERSION/$BUILD_DIR/
-UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$BUILD_DIR"
+
+if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
+  FOLDER=pull_requests
+else
+  if [ "$TRAVIS_BRANCH" == "master" ]; then
+    FOLDER=master
+  else
+    FOLDER=misc
+  fi
+fi
+
+BUILD_DIR="UTC-$(date -u +%T)_TRAVIS-$(echo $TRAVIS_BUILD_NUMBER)_COMMIT-$(echo $BUILD_VERSION)"
+
+mkdir -p uploads/$BUILD_DIR
+
+UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$FOLDER/"
+
+cp "./buildroot/output/images/boot.bin" "./uploads/$BUILD_DIR"
+cp "./buildroot/output/images/u-boot.img" "./uploads/$BUILD_DIR"
+cp "./buildroot/output/images/piksiv3.dtb" "./uploads/$BUILD_DIR"
+cp "./buildroot/output/images/zImage" "./uploads/$BUILD_DIR"
 
 echo "Uploading images to $UPLOAD_DIR"
-aws s3 cp "./output/images/boot.bin" "$UPLOAD_DIR"
-aws s3 cp "./output/images/u-boot.img" "$UPLOAD_DIR"
-aws s3 cp "./output/images/piksiv3.dtb" "$UPLOAD_DIR"
-aws s3 cp "./output/images/zImage" "$UPLOAD_DIR"
+aws s3 cp --recursive --no-sign-request ./uploads/ $UPLOAD_DIR
--- a/upload_artifacts.sh
+++ b/upload_artifacts.sh
@@ -10,12 +10,12 @@ BUILD_VERSION=$(git describe --tags --dirty --always)
 
 FOLDER=$TRAVIS_BRANCH
 if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
-  $FOLDER="pull_requests/$FOLDER"
+  FOLDER="pull_requests/$FOLDER"
 fi
 
 BUILD_DIR="$(date -u +%Y-%m-%dT%H:%M:%SZ)_${TRAVIS_BUILD_NUMBER}_${BUILD_VERSION}"
 
-UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$FOLDER/"
+UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$FOLDER/$BUILD_DIR/"
 
 FILES="boot.bin u-boot.img piksiv3.dtb zImage"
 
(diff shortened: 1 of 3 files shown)


upload_artifacts.sh
```diff
@@ -1,21 +1,33 @@
...
-set -e
+if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
+  FOLDER=pull_requests
+else
+  if [ "$TRAVIS_BRANCH" == "master" ]; then
+    FOLDER=master
+  else
+    FOLDER=misc
+  fi
+fi
```
mfine: this logic seems a little complex - how about something simpler:

```bash
FOLDER="$TRAVIS_BRANCH"
if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
FOLDER="pr/$FOLDER"
fi
```

That way there's a convention where the branch is always the folder, and there's an optional pr folder up front.

upload_artifacts.sh
```diff
@@ -1,21 +1,33 @@
...
 
-BUILD_VERSION=$(git rev-parse --short HEAD)
-BUILD_DIR=commit_$BUILD_VERSION
-# "folders" on S3 are prefix keys
-# http://stackoverflow.com/questions/33113354/how-to-create-a-folder-like-i-e-prefix-object-on-s3-using-the-aws-cli
-aws s3api put-object --bucket $BUCKET --key $PIKSI_VERSION/$BUILD_DIR/
-UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$BUILD_DIR"
+BUILD_DIR="UTC-$(date -u +%T)_TRAVIS-$(echo $TRAVIS_BUILD_NUMBER)_COMMIT-$(echo $BUILD_VERSION)"
```
mfine: Should use iso8601 for the date - don't need the prefixs for travs or commit - that should be self-evident:

```
BUILD_DIR="$(date -u +"%Y-%m-%dT%H:%M:%SZ")-$TRAVIS_BUILD_NUMBER-$BUILD_VERSION"
```

upload_artifacts.sh
```diff
@@ -1,21 +1,33 @@
...
+
+mkdir -p uploads/$BUILD_DIR
+
+UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$FOLDER/"
+
+cp "./buildroot/output/images/boot.bin" "./uploads/$BUILD_DIR"
```
mfine: why the `./uploads/`?

upload_artifacts.sh
```diff
@@ -1,21 +1,33 @@
 #!/bin/bash
 
-# Used in .travis.yml
+# This script is used in .travis.yml
+
+set -e
 
 BUCKET=piksi-buildroot-images
 PIKSI_VERSION=v3
+BUILD_VERSION=$(git rev-parse --short HEAD)
```
mfine: I think we consistently use:

```
git describe --tags --dirty --always
```

for the versions numbers. Should check.

margaret: Ah okay. I had just checked `git describe --tags` and that one said `fatal: No names found`

upload_artifacts.sh
margaret: Ahhhh you're right, I don't need that. For some reason I thought that uploading like `aws s3 cp --recursive yourfile.txt s3://bucket/prefixthatdoesntexistyet/` then that wouldn't work but it actually does 😑

upload_artifacts.sh
margaret: Won't that result in a gazillion folders? Or is having a gazillion folders okay?

upload_artifacts.sh
mfine: Ahh, has no tags. Will this repo have tags? If it does, we should follow the semantics for version in other repos so that it makes things consistent (don't have to map between multiple version schemes).

mfine: Could tag it with a `v0.0` to start :)

upload_artifacts.sh
mfine: every branch and master would get a folder in the root folder, and then there'd be a pr folder - shouldn't be too bad (might be worth making pr `PR` so that wouldn't conflict with a lower case named pr branch).

upload_artifacts.sh
```diff
@@ -8,14 +8,9 @@ BUCKET=piksi-buildroot-images
...
-else
-  if [ "$TRAVIS_BRANCH" == "master" ]; then
-    FOLDER=master
-  else
-    FOLDER=misc
-  fi
+  FOLDER="pull_requests/$FOLDER"
 fi
 
 BUILD_DIR="UTC-$(date -u +%Y-%m-%dT%H:%M:%SZ)_$(echo $TRAVIS_BUILD_NUMBER)_$(echo $BUILD_VERSION)"
```
mfine: Drop the 'UTC-'

upload_artifacts.sh
```diff
@@ -1,21 +1,26 @@
...
 PIKSI_VERSION=v3
+BUILD_VERSION=$(git describe --tags --dirty --always)
 
-set -e
+FOLDER=$TRAVIS_BRANCH
+if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
+  FOLDER="pull_requests/$FOLDER"
+fi
+
+BUILD_DIR="UTC-$(date -u +%Y-%m-%dT%H:%M:%SZ)_$(echo $TRAVIS_BUILD_NUMBER)_$(echo $BUILD_VERSION)"
```
mfine: also, don't need to echo the vars - just use them directly:

```
BUILD_DIR="$(date -u +%Y-%m-%dT%H:%M:%SZ)_$TRAVIS_BUILD_NUMBER_$BUILD_VERSION"

File: upload_artifacts.sh [hunk 7]
```diff
@@ -1,21 +1,26 @@
...
-aws s3api put-object --bucket $BUCKET --key $PIKSI_VERSION/$BUILD_DIR/
-UPLOAD_DIR="s3://$BUCKET/$PIKSI_VERSION/$BUILD_DIR"
+declare -a files=("boot.bin" "u-boot.img" "piksiv3.dtb" "zImage")
 
 echo "Uploading images to $UPLOAD_DIR"
-aws s3 cp "./output/images/boot.bin" "$UPLOAD_DIR"
-aws s3 cp "./output/images/u-boot.img" "$UPLOAD_DIR"
-aws s3 cp "./output/images/piksiv3.dtb" "$UPLOAD_DIR"
-aws s3 cp "./output/images/zImage" "$UPLOAD_DIR"
+for file in "${files[@]}"
```
mfine: should be able to do this without an array, which is more conventional:

```
FILES="boot.bin u-boot.img piksiv3.dtb zImage"
for file in $FILES
do
    echo $file
done
```

upload_artifacts.sh
margaret: had to do that because otherwise it globs the following `_` onto it and looks for `$TRAVIS_BUILD_NUMBER_`

```
$ export COMMIT=abcdefg
$ export TRAVIS_BUILD_NUMBER=42
$ echo "$TRAVIS_BUILD_NUMBER"; echo "$COMMIT"
42
abcdefg
$ echo "$TRAVIS_BUILD_NUMBER_$COMMIT"
abcdefg

```

mfine:
```bash
echo "${TRAVIS_BUILD_NUMBER}_${COMMIT}"
42_abcdefg
```

upload_artifacts.sh
```diff
@@ -10,7 +10,7 @@ BUILD_VERSION=$(git describe --tags --dirty --always)
 
 FOLDER=$TRAVIS_BRANCH
 if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
-  FOLDER="pull_requests/$FOLDER"
+  $FOLDER="pull_requests/$FOLDER"
```
mfine: That won't work. Needs to be as it was:

```
bash-3.2$ FOO="wut"
bash-3.2$ $FOO="yo/$FOO"
bash: wut=yo/wut: No such file or directory
bash-3.2$ FOO="yo/$FOO"
bash-3.2$ 
```

margaret:
@jacobmcnamee can I merge this after squashing the commits?

PR #29, opened 2016-10-04
Network boot support

jacobmcnamee:
TODO:
- [x] Update U-Boot pointer

/cc @gsmcmullin @mfine


gsmcmullin:
This is great.

For EVT2, I add the commands:
```
mdio write 9 0
mdio write 0 0
mdio write 0 1000
```
To force Ethernet to 100M and restart autonegotiation.

PR #30, opened 2016-10-04
SBP serial

gsmcmullin:
Turn of getty on serial port.
Enable USB serial gadget in kernel.
Run zmq adapter for serial ports on startup.


jacobmcnamee:
lgtm assuming it works

PR #37, opened 2016-10-28
Changes for production build on EVT2

gsmcmullin:


jacobmcnamee:
Note that this breaks compatibility with other QSPI flash chips

PR #42, opened 2016-11-22
Fix upload of EVT2 binaries

jacobmcnamee:
/cc @swift-nav/firmware


.travis.yml
```diff
@@ -19,7 +19,7 @@ after_success:
   - PRODUCT_VERSION=v3 PRODUCT_REV=evt1 PRODUCT_TYPE=dev ./publish.sh buildroot/output/images/piksiv3_evt1_dev/*
   - PRODUCT_VERSION=v3 PRODUCT_REV=evt1 PRODUCT_TYPE=prod ./publish.sh buildroot/output/images/piksiv3_evt1_prod/*
```
mfine: we should take `evt1` out.

PR #44, opened 2016-11-23
Support for building in Docker

fnoble:
**What:**
Buildroot only supports building under Linux. It would be helpful to have an easy way to build under OS X

**How:**
Using Docker we can define a repeatable build environment that can be used on any platform (including OS X). As a bonus, Docker is also supported by Travis allowing us to have consistency between our local and CI build environments.

The included Dockerfile sets up an Ubuntu linux container and installs the necessary dependencies before invoking the `build.sh` script.

**Testing:**
Tested locally on OS X as well as on Travis. S3 upload functionality on Travis and verified still working.

cc: @gsmcmullin @jacobmcnamee

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,23 @@
 sudo: required
-dist: trusty
+#dist: trusty
 
 language: c
 
+services:
+  - docker
+
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -y build-essential unzip
-  - sudo apt-get build-dep -y gcc
   - pip install --user --upgrade awscli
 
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,20 +1,23 @@
 sudo: required
-dist: trusty
+#dist: trusty
```
jacobmcnamee: Remove commented line?

jacobmcnamee:
Any particular reason to have Travis use Docker instead of native?

fnoble:
Incorporated into #48

PR #46, opened 2016-11-28
Piksi system daemon

gsmcmullin:
Add new `piksi_system_daemon`.  This new daemon process on Linux is responsible for implementing
system configuration settings for IPv4 Ethernet configuration and UART baudrates.

It also implements the new SBP shell command messages used for Piksi Multi firmware upgrade.  For firmware upgrade https://github.com/swift-nav/piksi_buildroot/pull/37 is also needed.

Do not merge until libsbp pointer has been updated:
https://github.com/swift-nav/libsbp/pull/412


package/piksi_system_daemon/piksi_system_daemon.mk
```diff
@@ -0,0 +1,20 @@
+################################################################################
+#
+# sbp_settings_client
```
jacobmcnamee: update name

package/piksi_system_daemon/src/main.c
```diff
@@ -0,0 +1,184 @@
...
+u8 eth_ip_mode = IP_CFG_STATIC;
+static char eth_ip_addr[16] = "192.168.0.222";
+static char eth_netmask[16] = "255.255.255.0";
+static char eth_gateway[16] = "192.168.0.1";
+
+static void eth_update_config(void)
+{
+  system("ifdown eth0");
+
+  FILE *interfaces = fopen("/etc/network/interfaces", "w");
```
jacobmcnamee: Does appending to the file every time the setting is written work as expected?

package/piksi_system_daemon/src/main.c
```diff
@@ -0,0 +1,184 @@
...
+  return i > 1 ? i : 0;
+}
+
+int shell_command_output(zloop_t *loop, zmq_pollitem_t *item, void *arg)
+{
+  struct shell_cmd_ctx *ctx = arg;
+  msg_log_t *msg = alloca(256);
+  msg->level = 6;
+
+  if (raw_fgets(msg->text, 254, ctx->pipe) > 0) {
```
jacobmcnamee: `sizeof(msg->text)` ?

jacobmcnamee:
Will this cause any problems for network boot in the dev config? Files are TFTP'd in another init script. Are the UARTs and network fully up and configured before the system daemon init script finishes?

package/piksi_system_daemon/src/main.c
gsmcmullin: `msg->text` is a zero length array.  Can do `SBP_FRAMING_MAX_PAYLOAD_SIZE - sizeof(*msg)` if that's better.

package/piksi_system_daemon/src/main.c
jacobmcnamee: Oh, yes it does. I guess as long as all the network interface settings (including possibly USB ethernet, for example) are contained in this daemon, we can just overwrite the whole file every time. Sounds fine.

gsmcmullin:
TFTP boot config still works for me.  The network interface is configured by the kernel in this case, and the files are transferred before the daemon is started.  It is reconfigured by the daemon to the values in the settings, but only after the firmware and bitstream have been downloaded.

jacobmcnamee:
Ok, sounds good. Is the default static IP also working?

gsmcmullin:
Yes, the default configuration is static 192.168.0.222/24, and works.

jacobmcnamee:
lgtm

PR #51, opened 2016-12-02
Turn off serial console output

jacobmcnamee:
To re-enable:
`env set bootargs console=ttyPS1,115200`

TODO:
- [ ] Update U-Boot: https://github.com/swift-nav/u-boot-xlnx/pull/19

/cc @swift-nav/firmware

### DIFF
--- a/board/piksiv3/piksiv3_microzed.dts
+++ b/board/piksiv3/piksiv3_microzed.dts
@@ -29,7 +29,6 @@
 
   chosen {
     bootargs = "clk_ignore_unused";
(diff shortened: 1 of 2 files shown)


fnoble:
is it possible to turn it off for prod and leave it on for dev?

fnoble:
@jacobmcnamee : looks like uboot has support for silencing output as a config option, this would let us set it only on the prod config.

https://github.com/swift-nav/u-boot-xlnx/blob/swift-v2016.1/doc/README.silent

PR #54, opened 2016-12-03
Fix firmware TFTP load from OS X tftpd

fnoble:
OS X tftpd doesn't like the absolute path. The linux one should work with either.

cc: @jacobmcnamee


anonymous-travis-swiftnav:
## 05be284
+ [s3://swiftnav-artifacts-pull-requests/piksi_buildroot/05be284](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/05be284/)

jacobmcnamee:
Might be good to test this on HITL, but I would be surprised if it didn't work. U-Boot already uses a relative path as far as I know.

fnoble:
@mfine and I tested manually by connecting a TFTP client to the HITL server. Works fine with the relative path.

PR #79, opened 2017-01-06
Add sleep to upgrade init script, don't delete file or reboot afterwards

denniszollo:
/cc @gsmcmullin @jacobmcnamee 

This works reliably on prod for me after adding the sleep.

- I changed the default warn level to warn (as that is the default level on the console and I wouldn't want to hide the messages)
- I also got rid of deleting the file and rebooting as it wasn't very user friendly.  This way, the user knows things were completed and isn't left guessing what happened.  Also the user would be able to take the flash drive and update another device afterwards.
- I lastly echoed the real name of the file

In the future we would like to:
- Blink PV led during upgrade
- Blink PV led afterwards differently
- make the serial output useful to someone in a plain text terminal as well as piksi console (maybe add some line breaks or something)
- Compare software versions on the file (from the header) and do not upgrade if it is identical to current software version

### DIFF
--- a/board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
+++ b/board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
@@ -1,10 +1,22 @@
 #!/bin/sh
 
+_mount() {
+    [[ $# -lt 3 ]] && { 
+        echo "Usage: ${FUNCNAME} <timeout_ms> <dev> <mnt_point>"; return 1
+    }
+    repeats="${1}"
+    until mount "${2} ${3}" 2>/dev/null || [[ $repeats -lt 1 ]]; do
+       $(( repeats-- )) 
+       usleep 1000
+    done
+}
+
+
 start() {
   export FIRMWARE="/media/sda1/PiksiMulti-*.bin"
   export LOGLEVEL="--warn"
-  # Sleep briefly to allow the automounting of flashdrive
-  sleep 0.5 
+  # try and mount for up to 0.5 seconds or until success 
+  _mount 500 /dev/sda1 /media/sda1
   if [ `echo $FIRMWARE | wc -w` != '1' ]; then
     echo 'Upgrade requires exactly one firmware image' | sbp_log $LOGLEVEL
     exit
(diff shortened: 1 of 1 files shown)


board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
```diff
@@ -2,8 +2,9 @@
 
 start() {
   export FIRMWARE="/media/sda1/PiksiMulti-*.bin"
-  export LOGLEVEL="--info"
-
+  export LOGLEVEL="--warn"
+  # Sleep briefly to allow the automounting of flashdrive
+  sleep 0.5
```
mfine: Can you poll here instead for the mount to show up.

board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
```diff
@@ -13,12 +14,15 @@ start() {
     exit
   fi
 
-  echo "New firmware image set detected: $FIRMWARE" | sbp_log $LOGLEVEL
+  echo "New firmware image set detected: `ls $FIRMWARE`" | sbp_log $LOGLEVEL
```
mfine: use `echo $FIRMWARE` to be consistent with above?

denniszollo: echo doesn't expand the wildcard

board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
denniszollo: I can poll with an ultimate timeout.  But we cannot block here because the user may be intending to upgrade.

denniszollo:
@mfine - now I hammer a mount command for up to 0.5 seconds.  Same effect on boot time as before, but at least we try a few more times.  As we gain confidence we could reduce the delay.

board/piksiv3/rootfs-overlay/etc/init.d/S84upgrade_tool
denniszollo: now I see your point, changed to echo.

PR #78, opened 2017-01-06
NMEA mode setting for UARTs

fnoble:
cc: @gsmcmullin @jacobmcnamee

PR #84, opened 2017-01-10
SBP message whitelist settings

gsmcmullin:
For: https://github.com/swift-nav/firmware_team_planning/issues/198
- Use filters on all external interfaces
- Add automatic reloading of `zmq_adapter` SBP filter config files.
- Add setting in `piksi_system_daemon` to update SBP filter config

- [x] Rearrange settings to match https://github.com/swift-nav/piksi_buildroot/pull/78/files#diff-f06aeebc999edf4b6ee6a5b216849d17L185


package/zmq_adapter/src/filter_sbp.c
```diff
@@ -107,6 +122,19 @@ int filter_sbp_process(void *filter_sbp_state,
 {
   filter_sbp_state_t *s = (filter_sbp_state_t *)filter_sbp_state;
 
+  /* Reload config if changed */
+  char buf[4096] __attribute__ ((aligned(__alignof__(struct inotify_event))));
```
jacobmcnamee: Where did `4096` come from?

package/zmq_adapter/src/filter_sbp.c
```diff
@@ -41,6 +46,16 @@ static int process_rule(filter_sbp_rule_t *rule)
 void filter_sbp_init(void *filter_sbp_state, const char *filename)
 {
   filter_sbp_state_t *s = (filter_sbp_state_t *)filter_sbp_state;
+  s->config_file = filename;
```
jacobmcnamee: Copy filename to a local buffer instead of assuming it persists outside of this function.

package/zmq_adapter/src/filter_sbp.c
```diff
@@ -41,6 +46,16 @@ static int process_rule(filter_sbp_rule_t *rule)
...
+  filter_sbp_load_config(s);
+  s->config_inotify = inotify_init();
+  /* inotify_initl seems to be missing, so use fcntl to set non-blocking */
+  fcntl(s->config_inotify, F_SETFL, O_NONBLOCK);
+  inotify_add_watch(s->config_inotify, filename, IN_MODIFY);
```
jacobmcnamee: Would it be better to use `IN_CLOSE_WRITE`?

jacobmcnamee: `inotify_init1()`
1 not L

jacobmcnamee: Can we print something out if this stuff fails?

package/piksi_system_daemon/src/whitelists.c
```diff
@@ -0,0 +1,143 @@
...
+ *  - All messages are sent
+ * "65535"
+ *  - Only the heartbeat message is sent (message ID 65535)
+ * "1234,5678/2,3456/10"
+ *  - Message 1234 is sent at full rate
+ *  - Message 5678 is sent at half rate
+ *  - Message 3456 is sent at 1/10 rate
+ */
+char wl[PORT_MAX][256] ={
+  [PORT_UART0] = "68,72,73,74,65535",
```
jacobmcnamee: Maybe we should remove the config files in `board/piksiv3/rootfs-overlay/etc` and let `piksi_system_daemon` create them. Otherwise I expect someone will modify them trying to change the defaults.

PR #143, opened 2017-03-09
added service to broadcast UDP discovery packets along with sender ID

axlan:


axlan:
Added write up: https://swiftnav.hackpad.com/Piksi-Broadcast-Discovery-Z8EWY5kzQJt

PR #144, opened 2017-03-09
Added command for getting ifconfig sent to console

axlan:


axlan:
Added better iterated approach that avoids system calls and uses new SBP message

PR #179, opened 2017-04-04
Update sbp_rtcm3_bridge to use libpiksi

jacobmcnamee:


jacobmcnamee:
Incorporated into #186

PR #185, opened 2017-04-04
Cell modem support (release branch)

gsmcmullin:
Adds cell modem support to Piksi
- Kernel drivers enabled for ACM devices and PPP networking
- Add pppd to userland
- Add settings to piksi_system_daemon to configure parameters and launch pppd.

Tested in NZ with on local carrier 2 Degrees.  Other carriers may require configuration of the APN setting.
This is not useful in itself, it just establishes the network interface.

For the connection to be successful, the modem must be connected, powered on and have a sim card installed.

/cc @swift-nav/firmware


anonymous-travis-swiftnav:
## v1.1.2-5-geb2e798
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.2-5-geb2e798/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.2-5-geb2e798/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.2-5-geb2e798

axlan:
I don't have a lot of context for the actual modem config, but the code looks good. The one concern I have is how the user would figure out the serial port if it doesn't enumerate as the default (or if there are multiple USB serial connections through a switch) . Currently they'd need to SSH in and check, which is probably fine for this relatively small use case.

jdormoy:
Gareth - if we put this end to end with the network client - what all would you recommend we test  Here are scenarios I thought would make sense:

1.   Going in and out of network range (will it reconnect?)
4.   are we able to  log output with it in different frequency

cbeighley:
@gsmcmullin Is this totally ready to merge?

PR #216, opened 2017-04-19
Upload limits

mfine:
Limit traffic being uploaded to Skylark. Previously, all traffic from the device had been heading to Skylark :( Now only LLH positions and heartbeats are uploaded. Additionally, these are rate limited by the filters.

/cc @jacobmcnamee @denniszollo


cbeighley:
@mfine Don't you want to upload obs and maybe ephemerides?

cbeighley:
Ah this is a workaround for the time being for the broker hammering?

mfine:
@cbeighley yeah, we'd like to eventually get obs and ephemeris and others up to the broker - but currently we're not using any of them. We'll definitely bring them in, but for now we should leave them out.

cbeighley:
👍

PR #262, opened 2017-05-05
Move bucket to swiftnav-releases

mfine:
To support the swiftnav-releases bucket.

### DIFF
--- a/fetch_firmware.sh.m4
+++ b/fetch_firmware.sh.m4
@@ -18,8 +18,8 @@ set -xe
 FW_VERSION=${1:-M4_FW_VERSION}
 NAP_VERSION=${2:-M4_NAP_VERSION}
 
-FW_S3_PATH=s3://swiftnav-artifacts/piksi_firmware_private/$FW_VERSION/v3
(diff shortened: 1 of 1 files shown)


anonymous-travis-swiftnav:
## v1.1.0-branch-160-gf0198c0
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-160-gf0198c0/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-branch-160-gf0198c0/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-160-gf0198c0

cbeighley:
@mfine I was thinking we would try to download the artifacts from the `swiftnav-releases` bucket, and if that fails, then try to download them using creds from the `swiftnav-artifacts` bucket. That way it will work out of the box for external users but should also be seamless for internal users.

jacobmcnamee:
The m4 file is only used for releases, right? If so, lgtm.

PR #274, opened 2017-05-15
[DO NOT MERGE] Try reduce image size

gsmcmullin:
- Use uClibc instead of glibc
- Remove some unused Linux drivers


lbara:
@gsmcmullin why not merging this PR?
Also, do you have an idea how much smaller the image would get?

PR #278, opened 2017-05-18
Volatile static flag to check before gps time gets updated.

mfine:
Don't decode until we get the callback. Not sure this is a good approach for a global with a callback.

/cc @anth-cole @pgrgich @gsmcmullin @jacobmcnamee


anth-cole:
this has been merged into master, and not needed on release now. Closing

PR #285, opened 2017-06-01
make ifup run in background to prevent blocking setting callback

axlan:
Addresses https://github.com/swift-nav/piksi_v3_bug_tracking/issues/407

The issue is the ifup command triggers Udhcpc which blocks the system call

### DIFF
--- a/package/piksi_system_daemon/src/main.c
+++ b/package/piksi_system_daemon/src/main.c
@@ -178,7 +178,7 @@ static void eth_update_config(void)
   }
   fclose(interfaces);
 
-  system("ifup eth0");
+  system("ifup eth0 &");
 }
 
 static int eth_ip_mode_notify(void *context)


axlan:
Tested working off DEV configuration.

Setting no longer hangs, but still has issue https://github.com/swift-nav/piksi_v3_bug_tracking/issues/489

anonymous-travis-swiftnav:
## v1.1.0-branch-182-gb604568
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-182-gb604568/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-branch-182-gb604568/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-182-gb604568

jacobmcnamee:
What will happen if that function is executed again before the running `ifup` completes?

axlan:
I can't confirm exactly what's happening, but immediately switching seems to end up with the second network configuration being correctly applied.

jacobmcnamee:
I think `ifup` and `ifdown` use a lockfile internally. So subsequent calls to either will block. I don't think this will actually completely fix the problem.

axlan:
I can't reproduce the problem on this branch even if I toggle back and forth as fast as the piksi console allows. I think there's a wider question of how we want to manage networking going forward (especially to allow wifi and ethernet dongles as well as auto-discovery), but I think this is a sufficient change to fix this bug for now.

jacobmcnamee:
Ok. I agree this should be better, just not a complete fix.

PR #291, opened 2017-06-29
Usb serial reopen fd on fail

lbara:
This fixes
https://github.com/swift-nav/firmware_team_planning/issues/332

I bet there are many better ways to fix this problem... comments more than welcome.

When unplugging/replugging the USB serial port, ttyGS0 produces EIO on write until it's closed and reopened.


package/zmq_adapter/src/zmq_adapter.h
```diff
@@ -19,10 +19,12 @@
 
 #include <czmq.h>
 
-void debug_printf(const char *msg, ...);
 void io_loop_start(int read_fd, int write_fd);
 void io_loop_wait(void);
 void io_loop_wait_one(void);
 void io_loop_terminate(void);
 
+#define debug_printf(format, ...) \
```
lbara: I created a macro so that I could actually add easily the calling function. There was no easy way with the static function and there were duplicated strings that were confusing (like "io loop begin", etc.)

anonymous-travis-swiftnav:
## v1.1.0-branch-209-g7f941ba
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-209-g7f941ba/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-branch-209-g7f941ba/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-branch-209-g7f941ba

gsmcmullin:
This prevents the child that writes to the port from dying, and output works after the cable is reconnected, but the process that reads still exits, and input no longer works.

Before disconnecting:
```
root@piksi:~ # ps aux | grep GS0
  997 root     zmq_adapter --file /dev/ttyGS0 -f sbp --filter-out sbp --filter-out-config /etc/usb0_filter_out_config -p >tcp://127.0.0.1:43031 -s >tcp://127.0.0.1:43030
 1027 root     zmq_adapter --file /dev/ttyGS0 -f sbp --filter-out sbp --filter-out-config /etc/usb0_filter_out_config -p >tcp://127.0.0.1:43031 -s >tcp://127.0.0.1:43030
 1028 root     zmq_adapter --file /dev/ttyGS0 -f sbp --filter-out sbp --filter-out-config /etc/usb0_filter_out_config -p >tcp://127.0.0.1:43031 -s >tcp://127.0.0.1:43030
 1122 root     grep GS0
```
After disconnecting and reconnecting:
```
root@piksi:~ # ps aux | grep GS0
  997 root     zmq_adapter --file /dev/ttyGS0 -f sbp --filter-out sbp --filter-out-config /etc/usb0_filter_out_config -p >tcp://127.0.0.1:43031 -s >tcp://127.0.0.1:43030
 1028 root     zmq_adapter --file /dev/ttyGS0 -f sbp --filter-out sbp --filter-out-config /etc/usb0_filter_out_config -p >tcp://127.0.0.1:43031 -s >tcp://127.0.0.1:43030
 1124 root     grep GS0
```

My approach to this would be to leave `zmq_adapter` as it was, and functionality to `piksi_system_daemon` to respawn `zmq_adapter` if/when it dies.

lbara:
So maybe instead of exiting cleanly we can exit with an error code

PR #303, opened 2017-07-14
Enabled Linux core dumps and send an sbp message when one occurs

lbara:


mookerji:
I'm not a reviewer, but this is **sweet**.

lbara:
This will do 3 things:
1) put a line in the syslog
2) send an sbp log with error severity
3) save a file in /tmp/cores

In the release build, I suggest that we check if the core directory exists (now it exists by default); if that folder exists, we save the log.

In other words, in the final release I still believe we should do 1) and 2), and 3) should be conditional. In that case, just creating the folder will "enable" the creation of the core dumps.

board/piksiv3/rootfs-overlay/usr/bin/dump_core.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+/usr/bin/logger "CORE DUMP!! executable=$1 pid=$2 time=$3"
+echo "CORE DUMP!! executable=$1 pid=$2 time=$3" | /usr/bin/sbp_log --error
+
+corefile="/tmp/cores/core-$1-$2-$3"
+touch $corefile
+
+while read line
+do
```
denniszollo: won't this mean we will fill up the ramfs completely rather than leaving some left over to allow the sytem to operate?

lbara: The core dumps have a mask where you say what sections to dump. Now with the mask 0x23 the core dumps are very tiny (less than 100 Kbytes).

Also, I set the core dump size to "unlimited", we can actually say 1 Mbyte or so.

denniszollo:
This looks awesome.
One request:
Is there a way we could save state during a session in order to count the number of times the process has crashed and put that in the log_error? Perhaps increment a number in a file?

This way from the SBP_LOG you will know if a crash happend prior to your logging or only after you started logging.  It might also help us catch crashes that happen before our logging systems or interfaces are up and running.

anonymous-travis-swiftnav:
## v1.1.0-develop-2017071417-4-gdad68e1
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017071417-4-gdad68e1/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-develop-2017071417-4-gdad68e1/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017071417-4-gdad68e1

board/piksiv3/rootfs-overlay/usr/bin/dump_core.sh
gsmcmullin: Apart from limiting the size of the core file, the cores could be dumped on a size limited tmpfs:
```
mount -t tmpfs -o size=16m none /tmp/cores
```

board/piksiv3/rootfs-overlay/usr/bin/dump_core.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+/usr/bin/logger "CORE DUMP!! executable=$1 pid=$2 time=$3"
+echo "CORE DUMP!! executable=$1 pid=$2 time=$3" | /usr/bin/sbp_log --error
+
+corefile="/tmp/cores/core-$1-$2-$3"
+touch $corefile
+
+while read line
```
gsmcmullin: I'm not crazy about this reading the core data as text and appending line-by-line.  I'd do this:
```
exec cat >> $corefile
```

lbara: I like your solution much better. I'll change it now

anonymous-travis-swiftnav:
## v1.1.0-develop-2017071417-5-ge9a1735
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017071417-5-ge9a1735/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-develop-2017071417-5-ge9a1735/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017071417-5-ge9a1735

PR #311, opened 2017-08-11
Can support

lbara:
Enabled CAN0 and CAN1.
They both come out as interfaces, but they are down.

I had to use iproute2 to make it work and added the can utils in the build.

Verified it works on Duro.


gsmcmullin:
Do we have any specification on what CAN support is actually required?
This exposes the interfaces in Linux, but we still won't have any actual support for using them.

Is there a problem with using ifconfig to bring up the interface?  My concern with adding unnecessary packages is that it increases the filesystem image size and therefore takes longer for the system to boot.

gsmcmullin:
Answering my own question here: It looks like iproute2 is the documented way to set the bit timing parameters.  Last time I looked at this it was done through a sysfs interface.

anonymous-travis-swiftnav:
## v1.1.0-develop-2017080422-4-gb506012
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017080422-4-gb506012/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-develop-2017080422-4-gb506012/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017080422-4-gb506012

PR #328, opened 2017-09-11
Kill any adapter running on GS0 when spawning a new one

lbara:
There is a bug, hard to reproduce but frequent enough, where multiple zmq_adapters are running on ttyGS0. This results in a bunch of duplicate messages.

For the time being, I think this code will take care of it.

I noticed the problem only once and could not figure out why that happened, but this is what I noticed:
* there were only the 2 forks alive, the main process had exited
* the 2 forks had ppid=1, which I guess was inherited from the spawning process

I was going down the route of using monit for checking those processes, but that's when I saw this problem, so I know for a fact that's not a fix.

Bear with me, this is not a great solution but I think it will fix our problem for 1.2

### DIFF
--- a/package/piksi_system_daemon/src/ports.c
+++ b/package/piksi_system_daemon/src/ports.c
@@ -181,8 +181,17 @@ static void adapter_kill(port_config_t *port_config)
 
 static int port_configure(port_config_t *port_config)
 {
+  /* kill adapter */
+  adapter_kill(port_config);
+
+  /* In case of USB adapters, sometimes we find that instances are still around.
+   * Kill them here
+   */
+  if (port_config->type == PORT_TYPE_USB) {
+    system("kill -9 `ps | grep GS0  | grep zmq_adapter | awk -F' ' '{print $1}'`");
+  }
+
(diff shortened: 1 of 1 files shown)


gsmcmullin:
It's not pretty, but if it's solves the problem let's merge for the release.  We should keep an open issue to come back to this to understand how it happened and fix the underlying cause.

anonymous-travis-swiftnav:
## v1.1.0-develop-2017090721-2-g9959c07
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017090721-2-g9959c07/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.1.0-develop-2017090721-2-g9959c07/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.1.0-develop-2017090721-2-g9959c07

PR #334, opened 2017-09-13
Crl/smooth pose integration

s-landers:
The crl/smooth-pose-integration branch on the s-landers fork of piksi_buildroot makes the necessary changes to buildroot to have it:

- Obtain a pose software archive from a git server hosted here at CRL
- Put the files from this archive in place in the target filesystem
- Create a service on the target system to automatically run the pose software


package/libsbp/libsbp.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-LIBSBP_VERSION = v2.2.0
```
lbara: Why do we need a different libsbp?

s-landers: There are libsbp messages this software sends that are present in this other version of libsbp that were not in v2.2.0.

PR #335, opened 2017-09-13
Pad new files, kernel and userspace buffers, fsck

silverjam:
Fixes https://github.com/swift-nav/piksi_v3_bug_tracking/issues/641

# Changes

+ We no longer use the 'sync' option in the auto mount script, this allows the kernel's "write back" system to optimize writes to block devices.

  Our current write back config is as follows:
  ```
  vm.dirty_background_ratio = 10
  vm.dirty_bytes = 0
  vm.dirty_expire_centisecs = 3000
  vm.dirty_ratio = 20
  ```

  Which means the write cache use up 20% of the system memory.  During my testing I saw that about 110mb of memory was available, so about 21mb of writes could potentially be cached.  We can tune these parameters to reduce the impact on SBP messages that are lost if a customer pulls out a removable media drive before a flush happens.

 + Use kernel and userspace buffers (revert changes to use open/write instead of fopen/fwrite from #168): this gets the slow usb drive up to about 300 kilobytes/s write throughput.  The fast SDHC card gets 5-7 megabytes/s.

  + Move all the time we spend updating the filesystem to one spot (when
    we create a new file).  For the typical case (a FAT fs on a thumb drive
    or sd card) this will give us a space to write data into that will not
    involve perturbing the FAT structures.  When the drive is ripped out the
    filesystem will not have a chance to be marked "clean" but there will be
    no FAT "cluster chains" in an unfinished state.

  + Auto fsck any FAT filesystem that's attached to the device to clean-up
     any lost data.  Lost data shows up as 'FSCK####.REC' files.

  + Force unmount so that the drive doesn't stay mounted when the
     card/drive is ripped out.

# Testing
Loaded new automount script and ran piksi at 20MHz solution frequency overnight.  This generates about 20kB/s of data.  Then used a [test script](https://gist.github.com/silverjam/7918b05d9b5112490e8d70cb8cae4214) to look for gaps.  Using [this script](https://gist.github.com/silverjam/243188f62b58b46a1a8f8c9fd5a7fe58) (and the associated `dd.c`) we can test the raw write speed of different block sizes for the media devices we're using on OS X.

# Addendum
We also drafted [a branch](https://github.com/swift-nav/piksi_buildroot/tree/hotfix/standalone_logger_write_thread) that moved writes onto a separate thread, and manually batched writes into large blocks.  This also fixes the issue, but probably requires significantly more QA than this change.


anonymous-travis-swiftnav:
## v1.2.0-branch-4-gbda32d5
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-branch-4-gbda32d5/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-branch-4-gbda32d5/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-branch-4-gbda32d5

anonymous-travis-swiftnav:
## v1.2.0-branch-4-gbda32d5
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-branch-4-gbda32d5/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-branch-4-gbda32d5/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-branch-4-gbda32d5

silverjam:
Test results after running overnight:
```
$ sbp2report=y ../gaps.py 0001.sbp
Max gap ms: 150, min gap ms: 100, num gaps: 3 (0.0004 %), missing: 27 (0.0036%), total: 746398
```

PR #346, opened 2017-09-25
Restart the system if piksi_system_daemon dies

lbara:
Refers to https://github.com/swift-nav/piksi_v3_bug_tracking/issues/701

### DIFF
--- a/board/piksiv3/rootfs-overlay/etc/monitrc
+++ b/board/piksiv3/rootfs-overlay/etc/monitrc
@@ -34,6 +34,7 @@ check process sbp_fileio_daemon_external with pidfile /var/run/sbp_fileio_daemon
 check process piksi_system_daemon with pidfile /var/run/piksi_system_daemon.pid
     start program = "/etc/init.d/S82piksi_system_daemon start"
     stop program = "/etc/init.d/S82piksi_system_daemon stop"
+    if 1 restarts within 3 cycles then exec "/sbin/reboot -f"
 
 check process sbp_rtcm3_bridge with pidfile /var/run/sbp_rtcm3_bridge.pid
     start program = "/etc/init.d/S82sbp_rtcm3_bridge start"


board/piksiv3/rootfs-overlay/etc/monitrc
```diff
@@ -34,6 +34,7 @@ check process sbp_fileio_daemon_external with pidfile /var/run/sbp_fileio_daemon
 check process piksi_system_daemon with pidfile /var/run/piksi_system_daemon.pid
     start program = "/etc/init.d/S82piksi_system_daemon start"
     stop program = "/etc/init.d/S82piksi_system_daemon stop"
+    if 1 restarts within 3 cycles then exec "/sbin/reboot -f"
```
silverjam: How long is a cycle?  This seems pretty nuclear.   Is there a way to kill everything and relaunch init instead of rebooting?  Would that be faster?

anonymous-travis-swiftnav:
## v1.2.3-4-gf83575c
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.3-4-gf83575c/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.3-4-gf83575c/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.3-4-gf83575c

board/piksiv3/rootfs-overlay/etc/monitrc
lbara: the problem is that when piksi_system_daemon restarts, new adapters are spawned, therefore we end up with a bunch of duplicate messages on each adapter (uart0, uart1, etc.)

The "cycles" is unclear to me, but since I wanted to reboot the system after the app dies, I figured that any number would work there (could be 1 restarts within 1 cycle and still do the same)

lbara: maybe there is a way to kill all the processes with ppid of the just-dead piksi_system_daemon. I'll look into that

PR #356, opened 2017-10-03
Update UART1 whitelist for usability.

denniszollo:
My task from triage meeting.

Requires console testing and SBP2RINEX testing before merging; both should work with a file recorded over UART1.

Will break HITL since HITL probably only uses the ECEF messages and is connected over UART1. @axlan

### DIFF
--- a/package/piksi_system_daemon/src/ports.c
+++ b/package/piksi_system_daemon/src/ports.c
@@ -78,7 +78,7 @@ typedef struct {
 static port_config_t port_configs[] = {
   {
     .name = "uart0",
-    .opts = "--file /dev/ttyPS0 --nonblock --outq 4096",
+    .opts = "--file /dev/ttyPS0 --nonblock --outq 8192",
     .opts_get = NULL,
     .type = PORT_TYPE_UART,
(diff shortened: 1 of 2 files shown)


package/piksi_system_daemon/src/ports.c
```diff
@@ -88,7 +88,7 @@ static port_config_t port_configs[] = {
   },
   {
     .name = "uart1",
-    .opts = "--file /dev/ttyPS1 --nonblock --outq 4096",
+    .opts = "--file /dev/ttyPS1 --nonblock --outq 4096 --debug",
```
mfine: Do we want a `--debug` flag checked in?

anth-cole:
lgtm - I'm nervous about checking in the `--debug` flag too, it seems like this could cause problems and maybe we should do this on master only.

denniszollo:
I think we will have to remove --debug flag before release.  

I also now propose that we double size of the serial out buffer queues so that in the worst case they hold 0.57 seconds worth of data rather than 0.28.

denniszollo:
I removed --debug flag (since it didn't seem to do much) and obs decimation.  I also tested extensively and feel good about it.

PR #366, opened 2017-10-09
Ignore SIGIO

lbara:
I have witnessed a situation where several adapters on UART0 were running.
Some of them were running as SBP, others as RTCM.

It appears that the proble is that the zmq_adapter exits when receiving SIGIO, which is unhandled.

### DIFF
--- a/package/piksi_system_daemon/src/ports.c
+++ b/package/piksi_system_daemon/src/ports.c
@@ -169,7 +169,7 @@ static void adapter_kill(port_config_t *port_config)
   sigchld_mask(&saved_mask);
   {
     if (port_config->adapter_pid > 0) {
-      int ret = kill(port_config->adapter_pid, SIGTERM);
+      int ret = kill(port_config->adapter_pid, SIGKILL);
       piksi_log(LOG_DEBUG,
                 "Killing zmq_adapter with PID: %d (kill returned %d, errno %d)",
                 port_config->adapter_pid, ret, errno);
--- a/package/piksi_system_daemon/src/ports.c
+++ b/package/piksi_system_daemon/src/ports.c
@@ -169,7 +169,7 @@ static void adapter_kill(port_config_t *port_config)
   sigchld_mask(&saved_mask);
   {
     if (port_config->adapter_pid > 0) {
-      int ret = kill(port_config->adapter_pid, SIGKILL);
+      int ret = kill(port_config->adapter_pid, SIGTERM);
       piksi_log(LOG_DEBUG,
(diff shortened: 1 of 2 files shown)


gsmcmullin:
The SIGTERM handler in zmq_adapter would kill its own children:
https://github.com/swift-nav/piksi_buildroot/blob/v1.2.0-release/package/zmq_adapter/src/zmq_adapter.c#L1111-L1120
https://github.com/swift-nav/piksi_buildroot/blob/v1.2.0-release/package/zmq_adapter/src/zmq_adapter.c#L326-L335

Sending SIGKILL will most likely leave the children running, and it's the children that are actually touching the port.

anonymous-travis-swiftnav:
## v1.2.11-2-g15c1c7b
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.11-2-g15c1c7b/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.11-2-g15c1c7b/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.11-2-g15c1c7b

anonymous-travis-swiftnav:
## v1.2.11-3-g4f4f0aa
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.11-3-g4f4f0aa/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.11-3-g4f4f0aa/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.11-3-g4f4f0aa

lbara:
I wasn't able to reproduce the issue after ignoring SIGIO in zmq_router.
This is how I tested it:

I ran manually zmq_adapter on UART0 with debug from command line many times. One out of 4 or 5 times I would get "I/O possible" on the console and the main thread would die, but not the write and read fork.

Investigating where that error comes from, I suspect that it started happening when we made the ports nonblocking. However, the common solution seems to be to ignore SIGIO, since this is not really an issue by itself.

lbara:
This should fix https://github.com/swift-nav/piksi_v3_bug_tracking/issues/789

PR #383, opened 2017-10-25
release_notes.txt: update version from v1.2.x to v1.2.14

denniszollo:


denniszollo:
Lets wait to merge until Thursday (just in case)

PR #403, opened 2017-11-30
WIP: Add a sample custom app

ljbade:
This adds a sample app for customers to use with the SDK.

A few things left to do:
- ~test it, need to run on my Piksi~
- disable it from building by default - will leave it out of the official firmware but make it easy for customer to enable in the buildroot config
- documentation, should link to https://geographiclib.sourceforge.io/cgi-bin/GeoidEval so people can customers up the geoid offset for their location

I am also not sure about the monitd/init scripts, we need them there to show the user how to get a daemon working, but at the same time it will produce errors when the app is disabled from compiling. Is it possible to get the rootfs files to be copied in as part of the package Makefile?

The other question, do we want to add a custom setting to the sample? Then the offset could be configured via the settings rather then hard coded in the init file.

Finally I copied `sbp.c` from the `sbp_rtcm3_bridge` package. For re-usability it might be a good idea to move this into `libpiksi`.

/cc @cbeighley @akshaybandiwdekar @silverjam


ljbade:
Just realised that the changes in `package/sbp_protocol/src/sbp_router.yml` should probably also only be made when the package is enabled. It is needed to filter out the original PVT message from the firmware and replace with the new one.

ljbade:
Implemented in https://github.com/swift-nav/piksi_buildroot/pull/419

PR #409, opened 2017-12-01
Allow UDP broadcast master

denniszollo:
Allow UDP broadcast on the client for master.

/cc @srossel @mfine @cbeighley

### DIFF
--- a/package/zmq_adapter/src/zmq_adapter_udp_connect.c
+++ b/package/zmq_adapter/src/zmq_adapter_udp_connect.c
@@ -38,12 +38,17 @@ static int addr_parse(const char *addr, struct sockaddr_in *s_addr)
 static int socket_create(const struct sockaddr_in *addr)
 {
   int ret;
-
+  int broadcastEnable = 1;
   int fd = socket(AF_INET, SOCK_DGRAM, 0);
   if (fd < 0) {
     return fd;
   }
-
+  
+  ret = setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcastEnable, sizeof(broadcastEnable));
+  if (ret != 0) {
+    goto err;
+  }
+  
   ret = connect(fd, (struct sockaddr *)addr, sizeof(*addr));
   if (ret != 0) {
     goto err;


anonymous-travis-swiftnav:
## v1.2.0-develop-2017113000-3-g98962c5
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017113000-3-g98962c5/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017113000-3-g98962c5/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017113000-3-g98962c5

mfine:
Might be more correct to do this only for an actual bcast address, and not always. E.g., if the address is a bcast per the network and mask (should be library routines to help check that).

Also should check out what needs to be done to support a mcast address. The receiver piece will likely need to join the mcast group to get data in.

/cc @silverjam

package/zmq_adapter/src/zmq_adapter_udp_connect.c
```diff
@@ -38,12 +38,15 @@ static int addr_parse(const char *addr, struct sockaddr_in *s_addr)
 static int socket_create(const struct sockaddr_in *addr)
 {
   int ret;
-
+  int broadcastEnable=1;
```
silverjam: Can we follow the existing style please?

```c
int broadcastEnable = 1;
```

package/zmq_adapter/src/zmq_adapter_udp_connect.c
```diff
@@ -38,12 +38,15 @@ static int addr_parse(const char *addr, struct sockaddr_in *s_addr)
...
   int fd = socket(AF_INET, SOCK_DGRAM, 0);
   if (fd < 0) {
     return fd;
   }
-
+  ret=setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcastEnable, sizeof(broadcastEnable));
+  if (ret != 0) {
+    goto err;
+  }
```
silverjam: Fix style please:
```c
ret = setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcastEnable, sizeof(broadcastEnable));
```

silverjam:
I think it's more work than it's worth to detect the subnet broadcast or the "local" broadcast address.  It looks like the "broadcast" flag is only added when then kernel detects that the address is 255.255.255.255.  And I don't think using a local subnet broadcast address would fall into this category, 255.255.255.255 is "dangerous" because it sends to all interfaces connected to the device.

The ipv4_is_lbcast function triggers the code that tags a packet as "broadcast", with ipv4_is_lbcast defined as (https://elixir.free-electrons.com/linux/latest/source/include/linux/in.h#L57, INADDR_BROADCAST is 0xFFFFFFFF):

```c
static inline bool ipv4_is_lbcast(__be32 addr)
{
	/* limited broadcast */
	return addr == htonl(INADDR_BROADCAST);
}
```

Which triggers code here:
  - https://elixir.free-electrons.com/linux/latest/source/net/ipv4/route.c#L2162
  - and here https://elixir.free-electrons.com/linux/latest/source/net/ipv4/route.c#L1987

Then, the flag is only checked if it's tagged as broadcast address:
https://elixir.free-electrons.com/linux/latest/source/net/ipv4/udp.c#L1029

```c
	if ((rt->rt_flags & RTCF_BROADCAST) && !sock_flag(sk, SOCK_BROADCAST)) {
		ip_rt_put(rt);
		err = -EACCES;
		goto out;
	}
```

denniszollo:
I updated the styling.  Sorry about that carelessness.

anonymous-travis-swiftnav:
## v1.2.0-develop-2017113000-4-gd2cad9c
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017113000-4-gd2cad9c/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017113000-4-gd2cad9c/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017113000-4-gd2cad9c

denniszollo:
@silverjam 

Does your comment above basically mean that my approach is okay?  Or do you think we need to revise?

silverjam:
@denniszollo Yeah, I think it's fine-- the only downside is that broadcasting to 255.255.255.255 might try to send data over the cell modem (if the device has one).  The cell network will more than likely ignore the broadcast.  You might try testing with 192.168.1.255 without setting the flag and see if it works.  If that works we can remove the flag altogether, and just instruct customers on how to calculate the broadcast address for their network.

denniszollo:
The reason I did this is because without the flag broadcasting to 192.168.1.255 didn't work.

PR #414, opened 2017-12-05
Auto-detect and log cell modem presence

gsmcmullin:
- Remove modem dev and type settings
- When new tty devices are found they are probed with AT command for
modem detection
- Modem make/model/revision/serial number are logged over SBP
- GSM/CDMA is autodetected so the correct connect script can be used

### DIFF
--- a/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
+++ b/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
@@ -57,6 +57,7 @@ static int modem_read(int fd, char *buf, size_t len, const char *const *response
     for(int i = 0; responses[i]; i++) {
       int rlen = strlen(responses[i]);
       if ((n >= rlen) && (strcmp(buf + n - rlen, responses[i]) == 0)) {
+        buf[n - rlen] = 0;
         return i;
       }
     }
@@ -76,22 +77,20 @@ static int cellmodem_command(int fd, const char *cmd, char *response, size_t len
     return -1;
 
   /* Check command echo */
-  char *tmp = strtok(buf, DELIM);
-  if ((tmp == NULL) || (strcmp(tmp, cmd) != 0))
+  if (strncmp(buf, cmd, strlen(cmd)) != 0)
     return -2;
 
   /* Pull out response */
   if (response) {
-    tmp = strtok(NULL, DELIM);
-    if (tmp == NULL)
-      return -3;
-    strncpy(response, tmp, len);
-  }
+    /* Strip \r\n junk */
+    int i;
+    for (i = strlen(buf) - 1; (buf[i] == '\r') || (buf[i] == '\n'); i--)
+      buf[i] = 0;
+    for (i = strlen(cmd); (buf[i] == '\r') || (buf[i] == '\n'); i++)
+      ;
 
-  /* Check for OK response */
-  tmp = strtok(NULL, DELIM);
-  if ((tmp == NULL) || (strcmp(tmp, "OK") != 0))
-    return -4;
+    strncpy(response, buf + i, len);
+  }
 
   return 0;
 }
--- a/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
+++ b/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
@@ -35,6 +35,11 @@ static const char *const response_strs[] = {
   NULL
 };
 
+/* Read a string from the modem, expected to end in one of the strings in
+ * `responses`.  `buf` must be large enough for the full response including
+ * the matched response.  A negative response indicates the expected string
+ * was not read, otherwise an index into `responses` is returned.
+ */
 static int modem_read(int fd, char *buf, size_t len, const char *const *responses)
 {
   int n = 0;
@@ -62,9 +67,14 @@ static int modem_read(int fd, char *buf, size_t len, const char *const *response
       }
     }
   } while (n < len - 1);
+  return -4;
 }
 
-#define DELIM "\r\n"
+/* Send a command and read back the response.  The modem is expected to reply
+ * with the echoed command, then the response, then `OK\r\n`.  When this
+ * pattern is matched, the response is returned in `response` and the return
+ * value is zero; returns negative otherwise.
+ */
 static int cellmodem_command(int fd, const char *cmd, char *response, size_t len)
 {
   /* Send command */
@@ -129,8 +139,10 @@ enum modem_type cellmodem_probe(const char *dev, sbp_zmq_pubsub_ctx_t *pubsub_ct
 
   /* Set to raw mode */
   struct termios tio = {0};
+  tcgetattr(fd, &tio);
   cfmakeraw(&tio);
-  cfsetspeed(&tio, B115200);
+  tio.c_cc[VMIN] = 0;
+  tio.c_cc[VTIME] = 0;
   if (tcsetattr(fd, TCSAFLUSH, &tio) < 0) {
     piksi_log(LOG_WARNING, "%s: tcsetattr failed on %s: %s", __func__, dev, strerror(errno));
     close(fd);
--- a/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
+++ b/package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
@@ -1,0 +1,170 @@
+/*
+ * Copyright (C) 2017 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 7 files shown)


anonymous-travis-swiftnav:
## v1.2.0-develop-2017120507-2-g86e576d
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-2-g86e576d/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017120507-2-g86e576d/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-2-g86e576d

gsmcmullin:
image

anonymous-travis-swiftnav:
## v1.2.0-develop-2017120507-3-g302b1fe
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-3-g302b1fe/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017120507-3-g302b1fe/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-3-g302b1fe

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+
+static int modem_read(int fd, char *buf, size_t len, const char *const *responses)
+{
+  int n = 0;
+  fd_set fds;
+  struct timeval timeout = {.tv_sec = 0, .tv_usec = 100000};
+  do {
+    FD_ZERO(&fds);
+    FD_SET(fd, &fds);
+    if (select(fd+1, &fds, NULL, NULL, &timeout) < 0)
```
silverjam: Shouldn't the first arg just be `1`?

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+      return -2;
+    if (!FD_ISSET(fd, &fds)) {
+      return -1;
+    }
+    int r = read(fd, buf + n, len - n - 1);
```
silverjam: So, buf is expected to be null terminated? And len is the expected to be the full length of the buffer `buf`?  Can we state those preconditions somewhere?

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+    if (r < 0) {
+        return -3;
+    }
+    n += r;
+    buf[n] = 0;
+    for(int i = 0; responses[i]; i++) {
+      int rlen = strlen(responses[i]);
+      if ((n >= rlen) && (strcmp(buf + n - rlen, responses[i]) == 0)) {
+        buf[n - rlen] = 0;
```
silverjam: Why do we need to NULL terminate again?  To clean-up the buffer?  Could this just be `buf[0]=0`?

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+    }
+    int r = read(fd, buf + n, len - n - 1);
+    if (r < 0) {
+        return -3;
+    }
+    n += r;
+    buf[n] = 0;
+    for(int i = 0; responses[i]; i++) {
+      int rlen = strlen(responses[i]);
+      if ((n >= rlen) && (strcmp(buf + n - rlen, responses[i]) == 0)) {
```
silverjam: This doesn't seem to work if the read returns something like `"OK\r"` then `"\n otherstuff"` because `buf + n - rlen` would put buffer pointer at start of `"uff"` in buffer, right?  Are there scenarios like this?

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+        buf[n - rlen] = 0;
+        return i;
+      }
+    }
+  } while (n < len - 1);
+}
+
+#define DELIM "\r\n"
```
silverjam: Is this used anywhere?

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+    for (i = strlen(cmd); (buf[i] == '\r') || (buf[i] == '\n'); i++)
+      ;
+
+    strncpy(response, buf + i, len);
+  }
+
+  return 0;
+}
+
+static void log_info(sbp_zmq_pubsub_ctx_t *pubsub_ctx, const char *fmt, ...)
```
silverjam: You can use the `sbp_log` helper for this.

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
```diff
@@ -0,0 +1,169 @@
...
+  }
+  if (!isatty(fd)) {
+    close(fd);
+    return MODEM_TYPE_INVALID;
+  }
+
+  /* Set to raw mode */
+  struct termios tio = {0};
+  cfmakeraw(&tio);
+  cfsetspeed(&tio, B115200);
```
silverjam: Is there anyway to add a constant for this?  We hard code the speed to 115200 for the pppd config file as well.  If there's every a possibility of it being different, it would be nice to have a constant.

gsmcmullin: These aren't real serial ports, at least not now, so the actual number is unimportant, but it must be something non-zero.  I think it's set to 9600 by default by the tty driver, which would be good enough.  Maybe I should just do `tcgetattr` and set the parameters I care about?  If we ever indent to support modems where this matters it would most likely need to become a setting.

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
gsmcmullin: Thanks, I'll do that!

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
gsmcmullin: Not anymore.  I'll remove it

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
gsmcmullin: At this stage, no.  The OK or ERROR is the last thing to come back.  There are some other possible responses to some commands for example 'CONNECT' or 'NO CARRIER', but the variable part always comes first.  This works for the patterns of the currently supported modems.  We can always change it in future if we need to support other patterns.

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
gsmcmullin: I'll add a comment explaining.

package/piksi_system_daemon/piksi_system_daemon/src/cellmodem_probe.c
silverjam: I think either way is fine, but comment might be nice for people without a lot of exposure to pseudo-terminals (like me).  Why does it need to be non-zero?

anonymous-travis-swiftnav:
## v1.2.0-develop-2017120507-6-gffdd33a
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-6-gffdd33a/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017120507-6-gffdd33a/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017120507-6-gffdd33a

gsmcmullin:
@axlan Yes, the presence of those four log info message indicate the modem is connected and was successfully probed.  Nothing is logged if there is no modem, or probing fails.
Because there is an unpredictable delay in the enumeration of the modem on USB, but it should happen as part of the boot process.  The only way to force it is to disconnect the modem and reconnect it.
Allowing a manual trigger is problematic, because if configured the modem will already be in use.  It's not impossible, but will be disruptive if it's in use.

Do you want this changed, or can I merge as is?

axlan:
The factory fixture should be able to account for this behavior, so it's fine to merge as is.

PR #431, opened 2017-12-20
NMEA Sentence upload

mfine:
This is the basics mechanics behind getting NMEA GGA sentences uploaded with libcurl - it does a trick to force READFUNC's to be called, and then uses pausing to rate limit the GGA uploads. I used a static global like @silverjam did because everything's in a single process space. Would need to get a cached NMEA GGA value in there to read up from.

/cc @anth-cole @silverjam

### DIFF
--- a/package/libnetwork/src/libnetwork.c
+++ b/package/libnetwork/src/libnetwork.c
@@ -37,6 +37,9 @@ const double PIPE_WARN_THRESHOLD = 0.90;
 /** Min amount time between "pipe full" warning messages. */
 const double PIPE_WARN_SECS = 5;
 
+/** Min amount time between NMEA messages. */
+const double XFER_SECS = 5;
+
 typedef struct {
   int fd;
   bool debug;
@@ -53,6 +56,7 @@ static volatile bool shutdown_signaled = false;
 static volatile bool cycle_connection_signaled = false;
 
 static time_t last_pipe_warn_time = 0;
+static time_t last_xfer_time = 0;
 
 void libnetwork_shutdown()
 {
@@ -165,6 +169,19 @@ static size_t network_upload_read(char *buf, size_t size, size_t n, void *data)
   return -1;
 }
 
+static size_t network_upload_write(char *buf, size_t size, size_t n, void *data)
+{
+  time_t now = time(NULL);
+  if (now - last_xfer_time >= XFER_SECS) {
+    last_xfer_time = now;
+    // REPLACE THIS WITH A NMEA GGA cached value
+    size_t ret = sprintf(buf, "$GPGGA,231704.00,3745.7451670,N,12223.3432196,W,1,09,0.9,1.23,M,0.0,M,,*76\r\n");
+    return ret;
+  }
+
+  return CURL_READFUNC_PAUSE;
+}
+
 /**
  * Abort the connection if we make no progress for the last 30 intervals
  */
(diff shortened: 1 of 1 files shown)


anonymous-travis-swiftnav:
## v1.2.0-develop-2017122003-2-g61b5575
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575

anonymous-travis-swiftnav:
## v1.2.0-develop-2017122003-2-g61b5575
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.2.0-develop-2017122003-2-g61b5575

package/libnetwork/src/libnetwork.c
```diff
@@ -390,7 +412,10 @@ void ntrip_download(const network_config_t *config)
 
   struct curl_slist *chunk = ntrip_init(curl);
 
+  curl_easy_setopt(curl, CURLOPT_PUT,              1L);
+  curl_easy_setopt(curl, CURLOPT_CUSTOMREQUEST,    "GET");
   curl_easy_setopt(curl, CURLOPT_URL,              config->url);
+  curl_easy_setopt(curl, CURLOPT_READFUNCTION,     network_download_read);
```
anth-cole: Should this be `network_upload_write`? Not sure if I'm understanding how the GGA put actually gets triggered

mfine: In libcurl, the direction goes the other way. The read functions means "give me something to read to upload". The write functions mean "I'm going to write something I've downloaded". So here this means give NMEA sentences to read that will get uploaded. I'd keep the `network_download_` prefix associated with the download function here (even though, confusingly, the download is doing some upload as well).

silverjam:
This seems reasonable, but I'm assuming we need to implement some more hooks into the ntrip daemon so that it can receive the current position?

I think a difficult part of this implementation is going to be figuring out a reasonable IPC mechanism to allow the ntrip daemon to receive the data.  We can abuse signals again (maybe combined with shared memory) or we can try to integrate zmq and curl (with the more complicated curl interface).

Or, maybe the simplest solution is to have once of the Linux daemons that receives the POS_LLH message to write the current position to a file that the ntrip_daemon can read?

mfine:
@silverjam yeah, that's the issue. I was assuming to do something minimal like what you already did for the base position in the rtcm decoder. I wouldn't invest time and energy into doing something more complicated - like the multi curl interface.

anth-cole:
@silverjam @mfine I was thinking of putting in an sbp callback to listen for the POS_LLH message and then converting to the NMEA string, I couldn't see an easy way to listen for the NMEA GGA string being sent from the firmware, but I really don't have a good idea of how all this fits together, so open to suggestions

mfine:
@anth-cole should be able to listen to the NMEA output by getting messages from the nmea routing:

https://github.com/swift-nav/piksi_buildroot/blob/v1.3.0-release/package/nmea_protocol/src/nmea_router.yml

listen to outgoing messages and cache the GGA messages (simple strcmp against the sentence). Should work the same as listening for POS_LLH messages but without having to do nmea encoding.

silverjam:
Merged in #444, thanks @mfine

PR #473, opened 2018-01-24
Only do the "no internet" check if NTRIP/Skylark is enabled

silverjam:


silverjam:
Already approved on v1.4 branch

PR #474, opened 2018-01-25
Fix glo obs warning when not receiving base station obs

benjaminaltieri:
Glo obs warning was firing off even when not connected to a base station.

Added a check for base station observations that gates this warning which is set in the message callback if any obs msg with the forwarding sender id is received.

swift-nav/piksi_v3_bug_tracking#999

### DIFF
--- a/package/health_daemon/src/glo_obs_monitor.c
+++ b/package/health_daemon/src/glo_obs_monitor.c
@@ -48,7 +48,10 @@ static health_monitor_t *glo_obs_monitor;
 static struct glo_obs_ctx_s {
   bool glo_setting_read_resp;
   bool glonass_enabled;
-} glo_obs_ctx = { .glo_setting_read_resp = false, .glonass_enabled = false };
+  bool base_obs_found;
+} glo_obs_ctx = { .glo_setting_read_resp = false,
+                  .glonass_enabled = false,
+                  .base_obs_found = false };
 
 static void
 sbp_msg_read_resp_callback(u16 sender_id, u8 len, u8 msg_[], void *ctx)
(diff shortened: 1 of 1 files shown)


benjaminaltieri:
Base station obs check will reset every timer cycle. So it would fire in the case the user disconnects the base station within the 10s window before any glonass obs have registered but after at least one regular obs made it through. I have no idea how we would account for that case, assuming the user would understand that they disconnected the base station.

benjaminaltieri:
Also, I am assuming this sort of gate should also be performed by the glo bias monitor. Does that sounds right?

anth-cole:
yeah - good point. These warnings are tricky to get comprehensive, lost of edge cases. We should do this for the glo bias message too, not sure how we can do that though right? In this instance we get the message but there's no obs in the message, in the bias message case we want to trigger if we don't receive a bias message, but only if we have an active reference stream

anonymous-travis-swiftnav:
## v1.4.2-2-g7f384b6
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.2-2-g7f384b6/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.2-2-g7f384b6/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.2-2-g7f384b6

benjaminaltieri:
Agreed. I'm going to merge this for now and add the bias warning as an issue to be fixed separately.

benjaminaltieri:
#475

PR #478, opened 2018-01-30
Fix caching of led brightness

benjaminaltieri:
First approach at solving swift-nav/piksi_v3_bug_tracking#997
Since the stuck leds are caused by a cache mismatch when the
piksi_leds daemon registers a successful brightness write but
the actual led brightness is incorrect. All brightness values
are updated every time the daemon updates the led states.

This may not be the actual approach we want to follow but it is available for consideration.

### DIFF
--- a/package/piksi_leds/src/led_adp8866.c
+++ b/package/piksi_leds/src/led_adp8866.c
@@ -23,12 +23,15 @@
 
 #define LED_I2C_ADDR 0x27
 #define LED_I2C_TIMEOUT 1
+#define LED_I2C_WRITE_RETRY_COUNT 3
 
 #define LED_LEVEL_SET_VALUE 0x20
 
 static int led_i2c;
 
-static u8 brightness_cache[LED_ADP8866_LED_COUNT] = {1};
+#define LED_I2C_CACHE_INITIAL_BRIGHTNESS (0u)
+static led_adp8866_led_state_t led_state_cache[LED_ADP8866_LED_COUNT];
+static bool flush_led_state_cache = false;
 
 /*
  * Duro connected different outputs of the adp8866 driver to the RGB
@@ -82,7 +85,6 @@ static int i2c_read(u8 addr, u8 *data) {
  * \return 0 if the operation succeeded, error message otherwise.
  */
 static int i2c_write(u8 addr, u8 data) {
-  int ret = 0;
   u8 buf[2] = {addr, data};
   struct i2c_msg msgs[] = {{
     .addr = LED_I2C_ADDR,
@@ -90,8 +92,7 @@ static int i2c_write(u8 addr, u8 data) {
     .buf = buf,
   }};
   struct i2c_rdwr_ioctl_data transaction = {.msgs = msgs, .nmsgs = 1};
-  ret = ioctl(led_i2c, I2C_RDWR, &transaction);
-  return ret < 0;
+  return ioctl(led_i2c, I2C_RDWR, &transaction) < 0;
 }
 
 /** Verify the contents of the MFDVID register.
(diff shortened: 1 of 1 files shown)


silverjam:
@benjaminaltieri What is the downside of not caching?  Writing lots of redundant data over i2c?  Instead of removing the cache, can we just flush it every once in a while?

anonymous-travis-swiftnav:
## v1.4.2-4-g06262b6
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.2-4-g06262b6/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.2-4-g06262b6/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.2-4-g06262b6

benjaminaltieri:
@silverjam yes, the alternative would be to periodically write all values, or fix the actual write failures somehow so that the cache stays consistent. This PR represents the least complex fix, I think the middle ground would be desirable (i.e. clear cache periodically)

silverjam:
@benjaminaltieri Do we see errors when the write fails?  Can we just retry a few times?

benjaminaltieri:
@silverjam nope, checkout the code here: https://github.com/swift-nav/piksi_buildroot/blob/1f943c2a6ab230808dd3bbd8c39d7265f527fea6/package/piksi_leds/src/led_adp8866.c#L217-L226

This is supposed to guard against the behavior we are seeing, unless it is being interpreted incorrectly. I can do some more debugging to see if this is the case.

benjaminaltieri:
Ok so it turns out that error check was basically being ignored. Adding a check for timeout with a retry, then a break for an actual error when encountered.

anonymous-travis-swiftnav:
## v1.4.3-3-g2202080
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.3-3-g2202080/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.3-3-g2202080/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.3-3-g2202080

package/piksi_leds/src/led_adp8866.c
```diff
@@ -215,12 +219,19 @@ static bool leds_set(const led_adp8866_led_state_t *led_states,
...
+        if (errno == ETIMEDOUT) {
+          ret = (i2c_write(isc_route[led_state->led],
+                     (led_state->brightness << LED_ADP8866_ISCn_SCDn_Pos)) == 0);
+        }
+      }
+
+      if (ret) {
         /* Update cache */
         brightness_cache[led_state->led] = led_state->brightness;
+      } else {
```
silverjam: I know this is WIP, but if this fixes the issue we should probably add a error logging & a cache flush here-- maybe add a retry loop here for a max of 3 retries?

benjaminaltieri:
The desired behavior here is that the cache becomes the intention of the driver, as opposed to actually representing the physical state of the LEDs. In the case that we had an error or multiple timeouts we simply mark the cache for flush on the next write. Assumptions are 1) we update the leds fairly often so flush will not be delayed, 2) we can catch all instances when the intended led update failed (ioctl and errno work as expected)

anonymous-travis-swiftnav:
## v1.4.3-4-gae3b488
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.3-4-gae3b488/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.3-4-gae3b488/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.3-4-gae3b488

package/piksi_leds/src/led_adp8866.c
```diff
@@ -318,6 +352,11 @@ bool led_adp8866_leds_set(const led_adp8866_led_state_t *led_states,
   u32 modified_states_count =
       modified_states_get(led_states, led_states_count, modified_states);
 
+  if (flush_led_state_cache) {
+    led_cache_flush();
```
silverjam: Do we care if there's an error here?

benjaminaltieri: Possibly, but at this point I'm not exactly sure how we would want to handle that error in a meaningful way. Simply leaving the flush flag set didn't seem like the proper thing to do so I've left it as is for now.

benjaminaltieri:
Dev PR merged #486

PR #479, opened 2018-01-30
Add shared context module for glo health monitors

benjaminaltieri:
The same parameters are used for both glo obs and
bias monitoring so they have been merged into
a global context with access/modifier methods.
Effectively this makes the bias monitor slave
to the more 'active' obs monitor - relying on
its shorter timeout to keep current the base
station connection parameter and handle the
glonass enabled setting query/updates

swift-nav/firmware_team_planning#453

Related to:  swift-nav/piksi_v3_bug_tracking#999

### DIFF
--- a/package/health_daemon/src/glo_bias_monitor.c
+++ b/package/health_daemon/src/glo_bias_monitor.c
@@ -10,6 +10,20 @@
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
 
+/**
+ * \file glo_bias_monitor.c
+ * \brief GLONASS Bias Message Health Monitor
+ *
+ * When glonass acquisition is enabled, it is expected that glonass bias
+ * measurements will be received periodically. This monitor will track
+ * bias messages and alert after a specified time period if no biases are
+ * received. Will not alert when not connected to a base station (no base
+ * obs messages being received).
+ * \author Ben Altieri
+ * \version v1.4.0
+ * \date 2018-01-30
+ */
+
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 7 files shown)


package/health_daemon/src/glo_bias_monitor.c
```diff
@@ -22,56 +22,25 @@
...
 
 static health_monitor_t *glo_bias_monitor;
 
-static struct glo_bias_ctx_s {
-  bool glo_setting_read_resp;
-  bool glonass_enabled;
-} glo_bias_ctx = { .glo_setting_read_resp = false, .glonass_enabled = false };
-
-static void
-sbp_msg_read_resp_callback(u16 sender_id, u8 len, u8 msg_[], void *ctx)
```
silverjam: Not sure I understand why we can remove this.  Are we querying the same setting in another health monitor?

silverjam:
Travis CI builds failed on network error for lzo package, manually re-kicked.

package/health_daemon/src/glo_bias_monitor.c
benjaminaltieri: Yeah this is part of the short term fix, which is to create a shared context `glo_health_context` which is effectively managed by the glo obs monitor. I almost merged both modules into one but went with this approach to keep the logic separate.

package/health_daemon/src/glo_bias_monitor.c
```diff
@@ -22,56 +22,25 @@
...
       LOG_WARNING,
-      "Glonass Biases Msg Timeout - no biases msg received within %d sec window",
-      GNSS_BIAS_ALERT_RATE_LIMIT / 1000);
-  }
-  if (!glo_bias_ctx.glo_setting_read_resp) {
-    health_monitor_send_setting_read_request(
-      monitor,
-      SETTING_SECTION_ACQUISITION,
-      SETTING_GLONASS_ACQUISITION_ENABLED);
+      "Reference Glonass Biases Msg Timeout - no biases received from base station within %d sec window",
```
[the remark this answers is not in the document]
benjaminaltieri: Will fix, thanks.

benjaminaltieri:
Fixed GLONASS in warnings messages and throughout code. Also slipped in some docs in the health monitors to address swift-nav/firmware_team_planning#447

benjaminaltieri:
#481

PR #487, opened 2018-02-06
Add Warning when NTRIP Connection Ends with 200

benjaminaltieri:
If an ntrip connection finishes with response code
and no data received we assume that the endpoint
was expecting a GGA message in order to begin the
transmission of VRS corrections. This does not
cover the case where an RTCM message indicating
such a requirement is sent.

Summary of features added:

 * A response code data member and check function

 * The check function is executed on completion
   of a network request that did not have a curl
   error and had a non-zero response code

 * An implementation of this feature to warn when
   the ntrip download ends with 200 and no data

 * The warning will only be issued once as it is
   a large message and the network request may
   repeat many times with the same result

Addresses swift-nav/piksi_v3_bug_tracking#1012

### DIFF
--- a/package/libnetwork/src/libnetwork.c
+++ b/package/libnetwork/src/libnetwork.c
@@ -60,6 +60,9 @@ struct network_context_s {
   curl_off_t bytes_transfered; /**< Current number of bytes transferred */
   curl_off_t stall_count;      /**< Number of times the bytes transferred have not changed */
 
+  long response_code;          /**< Response Code from last transfer request */
+  int (*response_code_check)(network_context_t* ctx); /**< Optional function reference to act on response code*/
+
   volatile bool shutdown_signaled;
   volatile bool cycle_connection_signaled;
 
@@ -89,11 +92,16 @@ static network_context_t empty_context = {
   .stall_count = 0,
   .shutdown_signaled = false,
   .cycle_connection_signaled = false,
+  .response_code = 0,
+  .response_code_check = NULL,
   .node = NULL,
   .url = "",
 };
 
 #define NMEA_GGA_FILE "/var/run/nmea_GGA"
+
+#define HTTP_RESPONSE_CODE_OK (200L)
+#define NTRIP_DROPPED_CONNECTION_WARNING "Connection dropped with no data. This may be because this NTRIP caster expects an NMEA GGA string to be sent from the receiver. You can enable this through the ntrip.gga_period setting."
 
 void libnetwork_shutdown()
 {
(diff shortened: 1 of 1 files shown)


package/libnetwork/src/libnetwork.c
```diff
@@ -520,6 +537,10 @@ static void network_request(network_context_t* ctx, CURL *curl)
       if (response != 0) {
         sbp_log(LOG_INFO, "Network Request Response Code - (%d)", response);
         piksi_log(LOG_INFO, "curl request code %d", response);
+        ctx->response_code = response;
+        if (network_response_code_check(ctx)) {
+          return;
```
mfine: Why exit here? Is the connection CURL context not retryable? Won't exiting here just have the process restart by monit instead?

anonymous-travis-swiftnav:
## v1.4.4-2-ge6f7280
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.4-2-ge6f7280/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.4-2-ge6f7280/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.4-2-ge6f7280

package/libnetwork/src/libnetwork.c
benjaminaltieri: The current implementation doesn't exit here. The intent was to allow context specific logic to determine whether or not to exit based on a response code. I'll remove it for now since it seems like there is never a case where we would want to stop looping on easy_perform

mfine: Yeah, that sounds good - I can't think of something currently to make us exit, especially as we're in a daemon loop.

anonymous-travis-swiftnav:
## v1.4.4-3-g6ff763f
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.4-3-g6ff763f/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.4-3-g6ff763f/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.4-3-g6ff763f

silverjam:
🛳it

benjaminaltieri:
Dev PR #488

PR #514, opened 2018-02-23
Dump serial output buffer when it's full

silverjam:
Fix for https://github.com/swift-nav/piksi_v3_bug_tracking/issues/1069

At 57600 baud with a 20Hz solution frequency, Piksi produces a stream of
data that's about 5-7 KiB/s -- near the limit that the output queue can be
emptied at 57600 baud.

This causing settings reads (which are lots of small packets) to be lost
or arrive in between lots of solution messages.  The best thing we can
do here (short of implementing priority queuing and smart rate limiting)
is to issue a warning that we're getting behind, and dump the old data
that's stuck in the queue.

Note that [this change from the 1.2.x release][1] to bump the "output queue"
limit to 8192 effectively disabled the feature because the queue would never
grow beyond 4096 for a serial device (whose internal buffer is also 4096).

[1]: https://github.com/swift-nav/piksi_buildroot/commit/89efd8b5aaf1687ed7fe0a053397affbed62bfa2 

However dropping random bits of data when the queue got full would cause
continuous CRC errors since we apparently don't submit whole SBP packets
to the underlying FD in zmq_adapter.

PR #515, opened 2018-02-24
 Dump serial output buffer when it's full [v1.4] 

silverjam:
Fix for swift-nav/piksi_v3_bug_tracking#1069
Release 1.4.x version of #514 

---

At 57600 baud with a 20Hz solution frequency, Piksi produces a stream of
data that's about 5-7 KiB/s -- near the limit that the output queue can be
emptied at 57600 baud.

This causing settings reads (which are lots of small packets) to be lost
or arrive in between lots of solution messages.  The best thing we can
do here (short of implementing priority queuing and smart rate limiting)
is to issue a warning that we're getting behind, and dump the old data
that's stuck in the queue.

Note that [this change from the 1.2.x release][1] to bump the "output queue"
limit to 8192 effectively disabled the feature because the queue would never
grow beyond 4096 for a serial device (whose internal buffer is also 4096).

[1]: https://github.com/swift-nav/piksi_buildroot/commit/89efd8b5aaf1687ed7fe0a053397affbed62bfa2 

However dropping random bits of data when the queue got full would cause
continuous CRC errors since we apparently don't submit whole SBP packets
to the underlying FD in zmq_adapter.

### DIFF
--- a/package/piksi_system_daemon/piksi_system_daemon/src/ports.c
+++ b/package/piksi_system_daemon/piksi_system_daemon/src/ports.c
@@ -97,10 +97,14 @@ typedef struct {
   restart_type_t restart;
 } port_config_t;
 
+// This is the default for most serial devices, we need to drop and flush
+//   data when we get to this point to avoid sending partial SBP packets.
+#define SERIAL_XMIT_SIZE "4096"
+
 static port_config_t port_configs[] = {
   {
     .name = "uart0",
-    .opts = "--file /dev/ttyPS0 --nonblock --outq 8192",
+    .opts = "--file /dev/ttyPS0 --nonblock --outq " SERIAL_XMIT_SIZE,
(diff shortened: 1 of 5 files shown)


anonymous-travis-swiftnav:
## v1.4.7-6-g866d602
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.7-6-g866d602/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.4.7-6-g866d602/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.4.7-6-g866d602

PR #535, opened 2018-03-09
Network usage reporting via Network Daemon

benjaminaltieri:
The goal of this PR is to facilitate the periodic updating of network usage information via the newly created SBP_MSG_NETWORK_BANDWIDTH_USAGE sbp message and includes a few discrete changes that are captured somewhat atomically in the commits but are summarized here:

 * Update libsbp to the latest version: v2.3.9 and whitelist the new sbp message
 * Add uptime utility function for use as duration in usage structs
 * Refactor existing interface code to avoid re-implementation and add a strategy
    extracting usage information directly into usage structs with that refactored code
 * Create `network_daemon` as an independent zloop reactor driven process that
    periodically uses the interface query APIs to send usage update over SBP

Testing:
 * Regression testing using the previously implemented SBP_MSG_NETWORK_STATE_RESP 
    and the console shows validates the refactored network interface code.
 * Bench testing with a WIP implementation of a 'cell modem wizard' widget
    that is wired to receive and display SBP_MSG_NETWORK_BANDWIDTH_USAGE in the
    same fashion as the Network element in the System Monitor tab of the console was
   used to receive the outputs from the `network_daemon`.


anonymous-travis-swiftnav:
## v1.3.0-develop-2018030818-5-g0d7cc92
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-5-g0d7cc92/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.3.0-develop-2018030818-5-g0d7cc92/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-5-g0d7cc92

package/libpiksi/libpiksi/src/networking.c
```diff
@@ -0,0 +1,248 @@
...
+#include <ifaddrs.h>
+#include <linux/if_link.h>
+
+#include <libpiksi/util.h>
+#include <libpiksi/networking.h>
+
+typedef s32 (*interface_map_fn_t)(struct ifaddrs *ifa, void *userdata);
+
+// refactored from original implementation in piksi_system_daemon by @axlan
+void map_network_interfaces(interface_map_fn_t map_fn, void *userdata)
```
silverjam: Looks like this can be static?

silverjam: Looks like this can be static?

package/libpiksi/libpiksi/src/networking.c
```diff
@@ -0,0 +1,248 @@
...
+  if (usage_entries == NULL
+      || usage_entries_count == NULL
+      || *usage_entries_count >= usage_entries_size) {
+    return -1;
+  }
+
+  // guard conditions to skip
+  if (ifa->ifa_addr == NULL
+      || ifa->ifa_name == NULL
+      || strlen(ifa->ifa_name) >= sizeof(usage_entries[0].interface_name)) {
```
silverjam: When would this happen?  Is this a programming error?

package/libpiksi/libpiksi/src/networking.c
```diff
@@ -0,0 +1,248 @@
...
+  struct fill_usage_struct_s *fill_usage = (struct fill_usage_struct_s *)userdata;
+  network_usage_t *usage_entries = fill_usage->usage_entries;
+  const u8 usage_entries_size = fill_usage->usage_entries_size;
+  u8 *usage_entries_count = fill_usage->usage_entries_count;
+  u64 duration = fill_usage->duration;
+
+  if (usage_entries == NULL
+      || usage_entries_count == NULL
+      || *usage_entries_count >= usage_entries_size) {
+    return -1;
```
silverjam: When would this happen?  Is this a programming error?

board/piksiv3/rootfs-overlay/etc/init.d/S86network_daemon
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+name="network_daemon"
+#cmd="network_daemon --interface eth0"
```
silverjam: What's the use case for tracking a single interface?

package/libpiksi/libpiksi/src/networking.c
benjaminaltieri: Assuming you mean the length check, it's valid since we have a limited buffer for the destination.

board/piksiv3/rootfs-overlay/etc/init.d/S86network_daemon
benjaminaltieri: Initially I believe we are only interested in the cell modem usage, this would act as a crude filter. At this point I have the filtering being done at the client side but it's not clear to me which is a more optimal solution.

package/libpiksi/libpiksi/src/networking.c
benjaminaltieri: Yes good catch. I was considering the mapping portion for a public interface for use elsewhere if needed but it seems like an unnecessary abstraction at this point since you still have to interact with the ifaddr references in the mapped function.

anonymous-travis-swiftnav:
## v1.3.0-develop-2018030818-6-g30e07ad
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-6-g30e07ad/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.3.0-develop-2018030818-6-g30e07ad/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-6-g30e07ad

board/piksiv3/rootfs-overlay/etc/init.d/S86network_daemon
silverjam: I think filtering on the device side would make more sense at the moment (so we can reduce the amount of data coming out of device), maybe this could be configurable (remotely) later if we want stats on other interfaces.

package/libpiksi/libpiksi/src/networking.c
silverjam: I mean, if we expect this to happen, we should issue a warning, if we don't expect it to happen we should assert/exit if it happens.

benjaminaltieri: Since this subroutine is mapped over all the interfaces, returning 0 in this context means `continue`, and returning -1 breaks the mapping loop like a `break`. Are you suggesting that instead of skipping an interface with a name that is too long, that we should stop further interface iteration?

board/piksiv3/rootfs-overlay/etc/init.d/S86network_daemon
benjaminaltieri: Agreed, I can remove the line if that is more clear.

anonymous-travis-swiftnav:
## v1.3.0-develop-2018030818-8-gc3e5e8d
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-8-gc3e5e8d/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.3.0-develop-2018030818-8-gc3e5e8d/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-8-gc3e5e8d

package/libpiksi/libpiksi/src/networking.c
benjaminaltieri: null checks seem standard, the count of entries is limited to the allocated memory so this is basically checking to see if we are out of slots to fill in. -1 breaks the mapping iteration

anonymous-travis-swiftnav:
## v1.3.0-develop-2018030818-9-ge6371e4
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-9-ge6371e4/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.3.0-develop-2018030818-9-ge6371e4/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018030818-9-ge6371e4

package/libpiksi/libpiksi/src/networking.c
silverjam: Maybe I'm not understanding how the code is used, but I don't want anything to fail silently, this will probably never happen, but let's make a habit to drop a quick log message if it does.

anonymous-travis-swiftnav:
## v1.3.0-develop-2018031620-9-g4d52740
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018031620-9-g4d52740/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.3.0-develop-2018031620-9-g4d52740/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.3.0-develop-2018031620-9-g4d52740

PR #571, opened 2018-04-09
Changed routing to handle GNSS and INS data at the same time

s-landers:


board/piksiv3/rootfs-overlay/etc/init.d/S80zmq_router_sbp
```diff
@@ -4,7 +4,7 @@ name="zmq_router_sbp"
 dir="/"
 user=""
 
-if grep -q "smoothpose_enable=True" /persistent/config.ini; then
+if grep -q "smoothpose_mode=INS Only" /persistent/config.ini; then
```
silverjam: Is this using the `settings_type_register_enum` from libpiksi or is the setting type a string?  If it's the latter we should change this to use an enum instead of a string.  Otherwise this looks fine.

https://github.com/swift-nav/piksi_buildroot/blob/268041fe322febbc9a93e83a9a890e783ed6029c/package/libpiksi/libpiksi/include/libpiksi/settings.h#L118

silverjam:
Closing in favor of https://github.com/swift-nav/piksi_buildroot/pull/576

PR #578, opened 2018-04-12
Duro inertial led behavior [CRL]

benjaminaltieri:
This is only for testing. Do not merge. Real PR here #557


benjaminaltieri:
DIdn't realize these aren't hooked up for CI

PR #584, opened 2018-04-19
Fix usb_upgrade return code issue (DEVC-428)

denniszollo:
should fix DEVC-428 and provide more informative user output if the image is corrupt or similar.  

The crux of the issue was using $? to try and get upgrade_tool return code after piping through sbp_log, which gives you return code from sbp_log instead of upgrade_tool

The cut thing is a little hacky but I think it's better than other ways to have a lookup table of strings for return codes

### DIFF
--- a/board/piksiv3/rootfs-overlay/etc/init.d/usb_upgrade.sh
+++ b/board/piksiv3/rootfs-overlay/etc/init.d/usb_upgrade.sh
@@ -3,6 +3,13 @@
 export FIRMWARE="/media/sda*/PiksiMulti-*.bin"
 export LOGLEVEL="--warn"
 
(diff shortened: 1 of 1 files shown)


anonymous-travis-swiftnav:
## v1.5.6-2-ge56c053
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.5.6-2-ge56c053/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.5.6-2-ge56c053/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.5.6-2-ge56c053

PR #585, opened 2018-04-19
Fix usb_upgrade return code issue (DEVC-428)

denniszollo:
should fix DEVC-428 and provide more informative user output if the image is corrupt or similar.  

The crux of the issue was using $? to try and get upgrade_tool return code after piping through sbp_log, which gives you return code from sbp_log instead of upgrade_tool

The cut thing is a little hacky but I think it's better than other ways to have a lookup table of strings for return codes


nicolekelley:
release PR - https://github.com/swift-nav/piksi_buildroot/pull/584

silverjam:
This will work for now, but it's pretty brittle, I would prefer if we have the upgrade_tool just report this string directly.  It can use sbp_log (from libpiksi/logging.h) to send it over sbp as well.

PR #593, opened 2018-04-24
Updated libSBP version, changed setting name checked by ZMQ router daemon startup script

s-landers:


package/libsbp/libsbp.mk
```diff
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-LIBSBP_VERSION = v2.3.15
+LIBSBP_VERSION = master
```
silverjam: Is there a release we can use that includes the new messages?  If there isn't we can find someone to cut a new one, I'm reluctant to track master.

PR #597, opened 2018-04-25
Add an explicit sync call to copy_sys_logs [DEVC-585] [v1.5]

silverjam:
This ensures that data actually makes it to the disk in the event of a
power failure.  For DEVC-585 we were seeing that a new directory was not
getting created in the event of a power cycle.  A new directory was
getting created, it was just the same as the one from the previous boot
because the data was never sync'd to the device.

### DIFF
--- a/board/piksiv3/rootfs-overlay/etc/init.d/copy_sys_logs.sh
+++ b/board/piksiv3/rootfs-overlay/etc/init.d/copy_sys_logs.sh
@@ -37,9 +37,11 @@ log_dir="$MOUNTPOINT/logs"
 mkdir -p "$log_dir"
 
 N=0
-while [[ -d "$log_dir/$N" ]] ; do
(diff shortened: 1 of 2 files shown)


anonymous-travis-swiftnav:
## v1.5.8-2-ged14afc
+ [Artifacts (HITL Dashboard Page)](http://hitl-dashboard.swiftnav.com/files/swiftnav-artifacts-pull-requests/piksi_buildroot/v1.5.8-2-ged14afc/)
+ [Artifacts (S3 / AWS Console)](https://console.aws.amazon.com/s3/home?region=us-west-2&bucket=swiftnav-artifacts-pull-requests&prefix=piksi_buildroot/v1.5.8-2-ged14afc/)
+ s3://swiftnav-artifacts-pull-requests/piksi_buildroot/v1.5.8-2-ged14afc

PR #639, opened 2018-05-28
[DEVC-322] Replace zloop usage in libpiksi with libuv

benjaminaltieri:
To accomplish the replacement of CZMQ usage throughout `piksi_buildroot`, a first step is to replace our main reactor, previously zloop, with a non GPLv3 library that can support similar interactions. The library chosen is [`libuv`](http://docs.libuv.org/en/v1.x/) which is a contemporary of `libevent` and `libev` primarily developed for use by Node.js.

Refactor highlights include:
* New loop module in `libpiksi` defining interaction of the new `pk_loop_t` context which wraps `libuv` usage under a generic API
* New endpoint module in `libpiksi` which wraps CZMQ `zsock_t` usage under a generic API (the next major step is to replace the guts of this with [`nanomsg`](https://github.com/nanomsg/nanomsg))
* Existing pubsub, tx, and rx contexts have been `s/zmq_//g`'d and refactored to use `pk_endpoint_t` contexts exclusively
* An 'attach' idiom has been introduced such that higher order contexts that hold rx endpoints can be given a `pk_loop_t` instance and own the process by which they manage addition and removal of endpoint readers (e.g. `settings_attach`, `sbp_rx_attach`)
* All packages previously running with `zmq_loop_simple` refactored to use `libpiksi` APIs
* Any usage of poll_item loops or other manual zmq operation has been replaced by the new or previously existing `libpiksi` APIs

Note: As of this submission the remaining packages `zmq_router` and `zmq_adapter` are to be refactored, as well as the addition of basic unit testing in `libpiksi` to cover these APIs. These are to be added or covered in a later PR

PR #644, opened 2018-06-05
[DEVC-322] Replace czmq with nanomsg

benjaminaltieri:
As a follow-up to #639, `czmq` has been completely removed from the codebase and replaced with [`nanomsg`](http://nanomsg.org/index.html). As `nanomsg` is a fork of `zmq`, the replacement was rather trivial and actually simplified much of the inner mechanics of the `endpoint` module. The commits have been split as clearly as possible to separate the functional changes from the cleanup steps.

High-level notes:
 * Since `czmq.h` had a lot of tag-along headers included, removing it meant having to chase down dependencies for many packages that ended up having missing includes.
 * Sockets created with `nn_socket` reveal a file descriptor that is level-triggered, as opposed to edge-triggered, so each `libuv` receive callback should correspond to a single message. The API has not changed but it was a detail that I had to comment on in the initial refactor that is slightly different now.
 * Our endpoint addresses have been revised to omit the '@' and '>' symbols that were used in `czmq` to indicate whether or not a socket should bind when created (vs connect). Since this was a unique and non-standard usage, I removed it in favor of explicitly creating *_SERVER versions of the PUB and SUB endpoint types. This has no impact on the current infrastructure since the only endpoints using bind were those created by the router so those are created as server types, while all others are unchanged in implementation and default to 'connect'. Practically, however, this removes the ability for an endpoint address to be trivially amended, in say a command line invocation, in order to affect these behaviors. From this point forward such options need to be routed into code, or hard coded into the process.


package/libpiksi/libpiksi/src/endpoint.c
```diff
@@ -100,196 +152,114 @@ pk_endpoint_type pk_endpoint_type_get(pk_endpoint_t *pk_ept)
...
-    int ret = zmsg_send(&msg, pk_ept->zsock);
-    if (ret == 0) {
+    int written = nn_send(pk_ept->nn_sock, data, length, 0);
+    if (written != -1) {
       /* Break on success */
-      break;
+      if (written == length) {
+        return 0;
+      } else {
+        piksi_log(LOG_ERR, "Message partial send (%d out of %d)", written, length);
```
silverjam: Can this be an assert? From reading the nanomsg docs this doesn't seem like something that will happen.

package/libpiksi/libpiksi/src/endpoint.c
```diff
@@ -100,196 +152,114 @@ pk_endpoint_type pk_endpoint_type_get(pk_endpoint_t *pk_ept)
...
-    piksi_log(LOG_ERR, "error in zmsg_addmem()");
-    zmsg_destroy(&msg);
-    return -1;
-  }
+  assert(pk_ept->type != PK_ENDPOINT_SUB || pk_ept->type != PK_ENDPOINT_SUB_SERVER);
 
   while (1) {
-    int ret = zmsg_send(&msg, pk_ept->zsock);
-    if (ret == 0) {
+    int written = nn_send(pk_ept->nn_sock, data, length, 0);
```
silverjam: Might be nice to expose the nn_allocmg  API through our wrapper so we can avoid some buffer copies.

package/libpiksi/libpiksi/include/libpiksi/endpoint.h
```diff
@@ -40,7 +40,9 @@ typedef struct pk_endpoint_s pk_endpoint_t;
  */
 typedef enum {
   PK_ENDPOINT_PUB,
+  PK_ENDPOINT_PUB_SERVER,
   PK_ENDPOINT_SUB,
+  PK_ENDPOINT_SUB_SERVER,
```
silverjam: If there are no usages of sub being a client (or pub being a server) I think we could make this the default  behavior for each and reduce the number of socket types. Alternately you could add a config enum/param to the endpoint creation function that would change the bind vs connect behavior. Yet another option that would be much more of a "drop in" change would be to duplicate the @ and > functionality (it could be simplified by making it a required part of the connection string).

package/libpiksi/libpiksi/src/endpoint.c
```diff
@@ -74,6 +97,16 @@ pk_endpoint_t * pk_endpoint_create(const char *endpoint, pk_endpoint_type type)
...
 
+  pk_ept->eid = do_bind ? nn_bind(pk_ept->nn_sock, endpoint)
+                        : nn_connect(pk_ept->nn_sock, endpoint);
+  if (pk_ept->eid < 0) {
+    piksi_log(LOG_ERR, "Failed to %s socket: %s",
+                       do_bind ? "bind" : "connect",
+                       pk_endpoint_strerror());
+    goto failure;
+  }
+  usleep(1000);
```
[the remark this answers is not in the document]
benjaminaltieri: Good catch, this was a debugging thing I forgot to remove.

package/libpiksi/libpiksi/src/logging.c
```diff
@@ -41,6 +47,12 @@ void piksi_log(int priority, const char *format, ...)
 
 void piksi_vlog(int priority, const char *format, va_list ap)
 {
+  if (debug_printf) {
+    char *with_return = (char *)malloc(strlen(format) + 1);
```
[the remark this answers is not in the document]
benjaminaltieri: Ouch, thanks for the catch!

package/libpiksi/libpiksi/src/endpoint.c
```diff
@@ -100,196 +152,114 @@ pk_endpoint_type pk_endpoint_type_get(pk_endpoint_t *pk_ept)
...
+  }
+  return recv_sock;
 }
 
 /**
- * @brief pk_endpoint_receive_zmsg - helper to retrieve a single message
+ * @brief pk_endpoint_receive_nn_msg - helper to retrieve a single message
  * @param pk_ept: pointer to the endpoint context
- * @return a newly allocated zmsg_t instance or NULL on failure
+ * @param buffer: pointer to existing buffer - OR - double pointer to receive allocated msg
```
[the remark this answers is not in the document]
benjaminaltieri: Yea since this is a static function I decided to just forward the nn_recv API behavior instead of implementing two versions of the same loop

package/libpiksi/libpiksi/include/libpiksi/endpoint.h
benjaminaltieri: We kind of talked about this 1-1, but basically all router socket use bind whether pub or sub, and other daemons use connect to communicate via the router. I really wanted to get rid of the non-standard '>' and '@' usage so I went with hardcoding our current usage. A larger tech debt refactor of the router might inform a better decision overall on this.

benjaminaltieri:
Updated from feedback, need to rebase on master once #639 is merged to get travis in on the fun

PR #658, opened 2018-06-13
Changed SBP router startup file to use new INS output setting name

s-landers:


jkretzmer:
Supporting settings changes in Pose Daemon.

PR #735, opened 2018-07-25
Galileo and Beidou on v1.5 buildroot

benjaminaltieri:
Attempting to test upgraded FPGA and FW images for v1.6 with the v1.5 release candidate.


benjaminaltieri:
No longer required as 2.0 has shipped

PR #742, opened 2018-07-25
Set the HTTP header after the init function

mfine:
Don't set the HTTP header in the init functions - set them after.

I noticed there are some instances where the chunk can get updated outside of the init function - not sure how this ends up picking the added headers. Move the setting of the header to outside of the init.

/cc @silverjam @benjaminaltieri


mfine:
But you have to keep updating that pointer on each add - it's a weird API.

silverjam:
Any reason this shouldn't be backported?  I created #743.

PR #774, opened 2018-08-14
Use SA_SIGINFO so we can know who sent a signal

silverjam:
+ Don't use `signal()` since the man page says not to use it.

+ Use SA_SIGACTION flag so that we can get info on who sent use a terminate signal

+ Initialize curl error buffer so we don't log garbage


silverjam:
Ported to master in #782

PR #793, opened 2018-08-30
Update upgrade_tool to v1.5.0-develop-2018083010

silverjam:
Bring in https://github.com/swift-nav/piksi_upgrade_tool/pull/4/files


silverjam:
Folded into #791

PR #833, opened 2018-09-14
Refactor sbp_rtcm3_bridge to rtcm3_to_sbp_bridge for consistentcy

anth-cole:
@benjaminaltieri - does this renaming make sense?


benjaminaltieri:
@anth-cole I think we actually went the opposite direction when we created `sbp_nmea_daemon`:
https://github.com/swift-nav/piksi_buildroot/blob/6603cd6b17997fdc1d26e7c91dee7961d3bd9161/configs/piksiv3_defconfig#L96
https://github.com/swift-nav/piksi_buildroot/blob/6603cd6b17997fdc1d26e7c91dee7961d3bd9161/Config.in#L33

plus didn't we want to do rtcm out at some point? would we have to create an `sbp_to_rtcm3_bridge` as a separate daemon?

anth-cole:
I think this is consistent with the nmea daemon isn't it? The first format is the origin format, and the second is the destination format.

This refactor is actually in anticipation of doing RTCM_OUT, and yeah, my plan was to do another daemon `sbp_to_rtcm3_bridge`

anth-cole:
after discussion with @benjaminaltieri we decided to bring the rtcm out capability within this daemon so the refactor is unnecessary

PR #850, opened 2018-09-20
revert Smoothpose pointer to 2.1.3

nicolekelley:
2.1.4 pointer did not have updated libnanomsg and is not working - need 2.1.5 fix from CRL


nicolekelley:
2.1.3 pointer has already been verified to work by @denniszollo

PR #915, opened 2018-10-08
Fix core dump on fileio [DEVC-1138]

yizow:
Attempting to list a directory that does not exist results in a
readdir(NULL) call. This causes a core dump.
We now ignore any requests on non-existing directories

PR #916, opened 2018-10-08
Fix core dump on fileio v2.1.0-release [DEVC-1138]

yizow:
Attempting to list a directory that does not exist results in a
readdir(NULL) call. This causes a core dump.
We now ignore any requests on non-existing directories


yizow:
See also:
https://github.com/swift-nav/piksi_buildroot/pull/915

PR #925, opened 2018-10-09
Fix memory error on libpiksi test [DEVC-1198]

yiiiyang2010:
There was a memory error causing the host build to crash on my machine.  Used valgrind to find the memory error and fix it.


silverjam:
Host build has passed, feel free to merge

PR #949, opened 2018-10-22
Bump Smoothpose pointer

denniszollo:


denniszollo:
@swiftnav-ram 

When this builds we can use it to test the initial velocity stuff.

PR #1022, opened 2018-11-27
reduce baseline alert to once every 10 seconds [v2.2]

silverjam:
(v2.2 version of #1019)

* reduce baseline alert to once every 10 seconds

* clang format


silverjam:
docker format passed

PR #1072, opened 2018-12-17
Fix creation of service dir links [ESD-968]

silverjam:
This prevents us from creating an "sv" link which points to nothing
and causes runsvdir to issue warnings about it.


silverjam:
Merging, last comment was just a comment addition (also killing travis).

PR #1189, opened 2019-03-22
Silverjam/sdk build

denniszollo:


board/piksiv3/post_image.sh
```diff
@@ -165,18 +167,32 @@ generate_failsafe()
...
+  local eas_tool_path=${HOST_DIR}/usr/bin/encrypt_and_sign
+  if [[ -e "$INTERNAL_BIN_PATH" ]]; then
+    if [[ -e $SDK_FPGA_SHA1 ]]; then
+      if [[ -f "$eas_tool_path" ]]; then
+        echo "ERROR: found 'encrypt_and_sign', for \"SDK\" builds, this tool not *SHOULD NOT BE* locatable ($eas_tool_path)" >&2
+        exit 1
+      fi
+      local sha=$(sha1sum "$FIRMWARE_DIR/piksi_fpga.bit" | cut -d' ' -f1)
+      if [[ "$sha" == "$(cat $SDK_FPGA_SHA1 | cut -d' ' -f1)" ]]; then
+        mv -v "$INTERNAL_BIN_PATH" "$SDK_BIN_PATH"
```
denniszollo: I think we might want to move this out of the post_image.sh script someday.  If you call make image-sdk, the title of the image is still internal, which is a little confusing.  The makefile could easily do the renaming.

silverjam: This step does happen as part of a makefile invocation.  You probably need to wipe out your `firmware/prod` directory so that you don't have any private artifacts laying-- the tricky part here is identifying when something is *really* an SDK build and no arbitrarily renaming something that actually has private artifacts (which is part of the danger of this build target, and highlights that we should always double check SDK builds before exposing them publicly or to third parties).

PR #1247, opened 2019-05-14
Add ethernet:interface_mode setting [ESD-1320]

yizow:
Adds a new ethernet:interface_mode setting to allow for batch updates of all
ethernet settings at once. This prevents the piksi from entering an
invalid IP state during the process of updating settings, and possibly
becoming unreachable.

If the ethernet configuration is in "Config" mode, the ethernet settings can be
changed freely, but will have no effect. When the ethernet configuration
is set to "Active", all settings will be updated at once, and become unmodifiable
until changed back to "Config" mode.

### DIFF
--- a/package/piksi_system_daemon/src/main.c
+++ b/package/piksi_system_daemon/src/main.c
@@ -61,7 +61,9 @@ static bool log_ping_activity = false;
 static const char const *ip_mode_enum_names[] = {"Static", "DHCP", NULL};
 enum { IP_CFG_STATIC, IP_CFG_DHCP };
 static u8 eth_ip_mode = IP_CFG_STATIC;
-static bool eth_locked;
+static const char const *interface_mode_enum_names[] = {"Config", "Active", NULL};
+enum { INTERFACE_MODE_ACTIVE, INTERFACE_MODE_CONFIG };
+static u8 interface_mode = INTERFACE_MODE_ACTIVE;
 static bool eth_settings_initialized = false;
 static char eth_ip_addr[16] = "192.168.0.222";
 static char eth_netmask[16] = "255.255.255.0";
@@ -88,7 +90,7 @@ static void eth_update_config(void)
 
 static settings_write_res_t eth_attempt_write(void)
 {
-  if (eth_settings_initialized && eth_locked) {
+  if (eth_settings_initialized && (interface_mode == INTERFACE_MODE_ACTIVE)) {
     sbp_log(LOG_WARNING, "Ethernet must be disabled to modify settings");
     return SETTINGS_WR_MODIFY_DISABLED;
   }
@@ -96,14 +98,14 @@ static settings_write_res_t eth_attempt_write(void)
   return SETTINGS_WR_OK;
 }
 
-static int eth_locked_notify(void *context)
+static int interface_mode_notify(void *context)
 {
   (void)context;
 
   /* Check if initialization is in process and this notify function was
    * triggered by read from persistent config file during boot. If this is the
    * case, other settings might not be ready yet. */
-  if (eth_settings_initialized && eth_locked) {
+  if (eth_settings_initialized && (interface_mode == INTERFACE_MODE_ACTIVE)) {
     eth_update_config();
   }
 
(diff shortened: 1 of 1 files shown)


package/piksi_system_daemon/src/main.c
```diff
@@ -61,6 +61,8 @@ static bool log_ping_activity = false;
 static const char const *ip_mode_enum_names[] = {"Static", "DHCP", NULL};
 enum { IP_CFG_STATIC, IP_CFG_DHCP };
 static u8 eth_ip_mode = IP_CFG_STATIC;
+static bool eth_locked;
```
benjaminaltieri: as discussed this will become an enum type `mode` with either `active` or `config` as options. `active` subs for `true` and `config` subs for `false`

silverjam: I think `interface mode` makes sense too help disambiguate it from the `ip config mode` we already have.

package/piksi_system_daemon/src/main.c
```diff
@@ -580,6 +605,14 @@ int main(void)
 
   settings_type_t settings_type_ip_mode;
   pk_settings_register_enum(settings_ctx, ip_mode_enum_names, &settings_type_ip_mode);
+  pk_settings_register(settings_ctx,
```
silverjam: This should be registered last to avoid the interface "bouncing" during boot-up.  Settings values are populated in the order that they are registered, so in order to avoid re-configuring the interface 4 times (once for each config field) we can simply register this field last.

yizow: There is a flag `eth_settings_initialized` set to true at the end of the settings registration; initially false. While false, the interface is not reconfigured.

This way the `mode` settings appears as the first option.

package/piksi_system_daemon/src/main.c
```diff
@@ -84,13 +88,37 @@ static void eth_update_config(void)
   }
 }
 
-static int eth_ip_mode_notify(void *context)
+static settings_write_res_t eth_attempt_write(void)
+{
+  if (eth_settings_initialized && (interface_mode == INTERFACE_MODE_ACTIVE)) {
+    sbp_log(LOG_WARNING, "Ethernet must be disabled to modify settings");
```
silverjam:
```suggestion
    sbp_log(LOG_WARNING, "Ethernet \"interface mode\" must be set to \"Config\" in order to modify settings");
```

package/piksi_system_daemon/src/main.c
```diff
@@ -580,6 +607,16 @@ int main(void)
 
   settings_type_t settings_type_ip_mode;
   pk_settings_register_enum(settings_ctx, ip_mode_enum_names, &settings_type_ip_mode);
+  settings_type_t settings_type_interface_mode;
+  pk_settings_register_enum(settings_ctx, interface_mode_enum_names, &settings_type_interface_mode);
+  pk_settings_register(settings_ctx,
+                       "ethernet",
+                       "mode",
```
silverjam:
```suggestion
                       "interface_mode",
```

silverjam: Never mind this, GH was showing me an old diff for some reason.

silverjam:
Doh:
```diff
diff --git a/package/piksi_system_daemon/src/main.c b/package/piksi_system_daemon/src/main.c
index 6c74e6656..c67edd633 100644
--- a/package/piksi_system_daemon/src/main.c
+++ b/package/piksi_system_daemon/src/main.c
@@ -91,7 +91,8 @@ static void eth_update_config(void)
 static settings_write_res_t eth_attempt_write(void)
 {
   if (eth_settings_initialized && (interface_mode == INTERFACE_MODE_ACTIVE)) {
-    sbp_log(LOG_WARNING, "Ethernet \"interface mode\" must be set to \"Config\" in order to modify settings");
+    sbp_log(LOG_WARNING,
+            "Ethernet \"interface mode\" must be set to \"Config\" in order to modify settings");
     return SETTINGS_WR_MODIFY_DISABLED;
   }
 
@@ -651,7 +652,8 @@ int main(void)
                        &eth_gateway);
 
   eth_settings_initialized = true;
-  interface_mode = INTERFACE_MODE_ACTIVE; // in case this value was saved to persistent - clear it out
+  interface_mode =
+    INTERFACE_MODE_ACTIVE; // in case this value was saved to persistent - clear it out
   eth_update_config();
 
   settings_type_t settings_type_time_source;
```

benjaminaltieri:
previous HITL results look fine, merging..

PR #1266, opened 2019-05-22
Add a settings to disable the sample daemon by default [ESD-1341]

silverjam:
As part of the build variant / binary artifact support for piksi_buildroot we enabled the "sample daemon" in the internal build so that we don't have to have a whole separate "SDK" build just to verify that the sample daemon builds. This cause the sample daemon to show up in the internal build and start logging things.  We'll continue to include sample daemon in the internal build, but we fix this by adding a setting to disable the sample daemon operation under normal circumstances.

### DIFF
--- a/package/sample_daemon/src/sample_daemon.c
+++ b/package/sample_daemon/src/sample_daemon.c
@@ -35,6 +35,7 @@
 #define PROGRAM_NAME "sample_daemon"
 
 static double offset = 0;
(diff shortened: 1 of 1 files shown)


silverjam:
@benjaminaltieri k, done

silverjam:
Setting is showing up and sample daemon is no longer logging things:

Screen Shot 2019-05-22 at 1 50 38 PM

PR #1274, opened 2019-06-05
Add protocol blacklist to port mode selection [ESD-1289]

yizow:
Certain CAN protocols such as J1939 are only applicable to the CAN ports
on the piksi. We should not be allowed to set the USB, UDP/TCP, etc.
ports into J1939 mode.

This commit accomplishes that by introducing protocol blacklists to the
ports daemon. Each port_config_t has a list of strings of protocol names
that are blacklisted on that port.

A list of all protocols is stored in a "global" array. Each port needs
to maintain their own list due to potential blacklisted protocols. Two
additional fields have been added to port_config_t to store the mapping
between these two lists.

### DIFF
--- a/package/ports_daemon/ports_daemon/src/ports.c
+++ b/package/ports_daemon/ports_daemon/src/ports.c
@@ -605,9 +605,9 @@ static int mode_enum_names_get(const char ***mode_enum_names, port_config_t *por
       if (strcmp(port_config->blacklist[x], protocol->setting_name) == 0) {
         piksi_log(LOG_DEBUG,
                   "skipping blacklisted protocol - port: %s name: %s",
-                  port_config->blacklist[x],
+                  port_config->name,
                   protocol->setting_name);
-        port_config->blacklisted_indices[x] = 1;
+        port_config->blacklisted_indices[protocol_index] = 1;
         enum_names[protocol_index_to_mode(protocols_used)] = NULL;
         blacklisted = true;
       }
--- a/package/ports_daemon/ports_daemon/src/ports.c
+++ b/package/ports_daemon/ports_daemon/src/ports.c
@@ -27,6 +27,9 @@
 #include "ports.h"
 #include "protocols.h"
 
+#define MAX_NUM_PROTOCOLS 10
+#define MAX_PROTOCOL_NAME_LEN 10
+
 #define MODE_NAME_SBP "SBP"
 
 #define MODE_NAME_DEFAULT MODE_NAME_SBP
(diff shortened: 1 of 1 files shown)


package/ports_daemon/ports_daemon/src/ports.c
```diff
@@ -97,17 +97,21 @@ typedef struct port_config_s {
   pid_t adapter_pid; /* May be cleared by SIGCHLD handler */
   restart_type_t restart;
   bool first_start;
+  const int blacklist_len;
+  const char *const blacklist[10];
+  int mode_mapping[10];           /* maps from global mode list to port mode sublist */
```
silverjam: Would it be simpler to just use the global mode mapping list and just mark certain indexes as invalid?  This would allow us to not have to maintain this sublist mapping indexes->mode names.

silverjam:
@yizow Is there a "companion" change for this showing how it would be utilized?

package/ports_daemon/ports_daemon/src/ports.c
yizow: oh, thats a good idea.

yizow:
yes, https://github.com/swift-nav/piksi_buildroot/pull/1251 is a "companion" change
However, i haven't pushed it yet to avoid having a duplicate commit / needing to rebase after this PR gets merged into master.

yizow:
@silverjam See https://github.com/swift-nav/piksi_buildroot/pull/1251/commits/6b232de6ff09f83a3dc60410440e301c19039e3e for companion change.
Blacklisting a protocol involves simply incrementing the blacklist_len field and adding the the blacklist array

package/ports_daemon/ports_daemon/src/ports.c
```diff
@@ -364,6 +398,33 @@ static u8 protocol_index_to_mode(int protocol_index)
...
+    if (port_config->blacklisted_indices[i] == 1) {
+      blacklisted_counter++;
+    }
+  }
+
+  return protocol_index_to_mode(mode_index - blacklisted_counter);
+}
+
+static int blacklisted_index_lookup(int index, const port_config_t *port_config)
+{
```
silverjam: I have no idea what this is doing.  Is this trying to fix the offset of the index when going from blacklisted index back to a regular index?  Where the offset is dependent on the number blacklisted index that appear before that index?

yizow: Ben and I did a pair programming session where we went through and refactored the code.
This function was removed - as you noted, it wasn't really doing anything useful. 

The new commit message should (hopefully) explain the new implementation, which is a lot simpler.

benjaminaltieri:
FYI @yizow, I'm recusing myself from review since I coauthored the latest revision. Not sure how that should work but afaic this is good to go :shipit:

package/ports_daemon/ports_daemon/src/ports.c
```diff
@@ -97,17 +102,21 @@ typedef struct port_config_s {
   pid_t adapter_pid; /* May be cleared by SIGCHLD handler */
   restart_type_t restart;
   bool first_start;
+  const int blacklist_len;
```
benjaminaltieri: just realized this isn't strictly necessary but could use a macro (defined in so many ways throughout codebases) like this: `#define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))`

PR #1283, opened 2019-06-27
Use new gnss-converters tag [ESD-1393]

mbavaro:
the current gnss-converters code can generate a MSM message for QZSS but fails to handle it in https://github.com/swift-nav/gnss-converters/blob/master/c/src/sbp_rtcm3.c#L92

Additional cmake changes in all repos require submodules to be pulled in as well..

### DIFF
--- a/package/gnss_convertors/gnss_convertors.mk
+++ b/package/gnss_convertors/gnss_convertors.mk
@@ -4,9 +4,12 @@
 #
 ################################################################################
 
-GNSS_CONVERTORS_VERSION = v0.3.84
-GNSS_CONVERTORS_SITE = https://github.com/swift-nav/gnss-converters
+# Workaround a Jenkins issue, this is the hash for v0.3.85 but Jenkins
+#   has cached v0.3.85 incorrectly.
+GNSS_CONVERTORS_VERSION = v0.3.86
+GNSS_CONVERTORS_SITE = git@github.com:swift-nav/gnss-converters.git
 GNSS_CONVERTORS_SITE_METHOD = git
+GNSS_CONVERTORS_GIT_SUBMODULES = YES
 GNSS_CONVERTORS_SUBDIR = c
 GNSS_CONVERTORS_DEPENDENCIES = libswiftnav librtcm libsbp
 GNSS_CONVERTORS_INSTALL_STAGING = YES
--- a/package/gnss_convertors/gnss_convertors.mk
+++ b/package/gnss_convertors/gnss_convertors.mk
@@ -4,10 +4,8 @@
 #
 ################################################################################
 
-# Workaround a Jenkins issue, this is the hash for v0.3.85 but Jenkins
-#   has cached v0.3.85 incorrectly.
 GNSS_CONVERTORS_VERSION = v0.3.86
-GNSS_CONVERTORS_SITE = git@github.com:swift-nav/gnss-converters.git
+GNSS_CONVERTORS_SITE = https://github.com/swift-nav/gnss-converters
 GNSS_CONVERTORS_SITE_METHOD = git
 GNSS_CONVERTORS_GIT_SUBMODULES = YES
 GNSS_CONVERTORS_SUBDIR = c
--- a/package/librtcm/librtcm.mk
+++ b/package/librtcm/librtcm.mk
@@ -5,7 +5,7 @@
 ################################################################################
 
 LIBRTCM_VERSION = v0.2.43
-LIBRTCM_SITE = git@github.com:swift-nav/librtcm.git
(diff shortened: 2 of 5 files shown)


mbavaro:
don't know what's happening..
according to `pkg-download.mk` and `pkg-generic.mk` simply adding `_GIT_SUBMODULES` should trigger a `-r` as
```
		$(if $($(PKG)_GIT_SUBMODULES),-r) \
```

yizow:
Unfortunately looks like the error is still present on hitl

yizow:
The firmware binaries from hitl all run fine on my desk piksi as well. Not sure what the difference between that and hitl would be...

mbavaro:
With this last commit, I am trying to update all gnss-converters submodules as per
https://github.com/swift-nav/gnss-converters/tree/master/c/third_party
Let's see if it helps.

mbavaro:
nothing:
https://hitl-viewer.ce.swiftnav.com/test_runs/metrics_preset=pass_fail&metric=Solution+Achieved+-+rtk_combined+Check&scenario=live-roof-leith-sym1-msm4&op=%3C%3D&threshold=0&build_type=buildroot_pr&firmware_versions=v2.3.0-develop-2019062617-10-g48f1e37c8
shows still the bridge failing to start

ericsenant:
Not sure my review is very relevant as I am not sure to clearly understand the real issue here

mbavaro:
@ericsenant no problem, it's just FYI as you pinged me on the Jira issue
@yizow with my tools I don't see the error anymore, could you please double check with your method?
@silverjam if it's acceptable to use hashes in master for buildroot, I'd proceed with these changes and close the Jira issue

anth-cole:
@mbavaro rather than pointing at the underlying hash, would it make sense to just cut a new release in these repos?

yizow:
Scheduling additional HITL runs to double check issue is gone

mbavaro:
@anth-cole I don't know.. these tags are pretty meaningless to me.. sure I would agree in release phase, but during master development it seems unjustified work more than anything else to me.. but if it has to be then I'll do it!

mbavaro:
@yizow bear in mind I scheduled runs in Australia (where QZSS is present) explicitly ;)

package/librtcm/librtcm.mk
```diff
@@ -4,9 +4,10 @@
 #
 ################################################################################
 
-LIBRTCM_VERSION = v0.2.41
-LIBRTCM_SITE = https://github.com/swift-nav/librtcm
+LIBRTCM_VERSION = 8f6c945f137a572716f7a4a2a75797260ac1e280
```
benjaminaltieri: let's do this properly and cut a release. there's no rush to fix this and unnecessary tech debt doesn't seem like the right call

yizow:
@mbavaro Ah, i see. 
Looks all clear on my end.

Incidentally, the one live-roof-townsend run that completed also passed with no errors.

anth-cole:
@mbavaro I find it really helpful to use the release tags as it's much easier (for me) to track what is included in the build.

package/gnss_convertors/gnss_convertors.mk
```diff
@@ -4,9 +4,12 @@
 #
 ################################################################################
 
-GNSS_CONVERTORS_VERSION = v0.3.84
-GNSS_CONVERTORS_SITE = https://github.com/swift-nav/gnss-converters
+# Workaround a Jenkins issue, this is the hash for v0.3.85 but Jenkins
```
benjaminaltieri: @schniedergers what is this issue? assuming the solution is to use `git@` instead of `https`, is this still needed?

silverjam: This is the jenkins+buidlroot caching issue, Jenkins permanently caches whatever buildroot downloads, then buildroot will refuse to retry to download if it thinks it has already grabbed the version specified.  In this case we didn't specify the "submodules" option for gnss-converters for version v0.3.85 (and we weren't forcing to use git to download), so it didn't grab the cmake submodule.  The change to use a hash was done just so that buildroot would be forced to download a "new" version with the correct download method (git + submodule fetch).  If we just bump the version from `v0.3.85` to `v0.3.86` we can remove this comment.

silverjam: Also, this comment is already out of date, since @mbavaro pushed a commit on top of `v0.3.85`.

benjaminaltieri: makes sense, thanks @silverjam. shouldn't be an issue moving forward now that these have been updated with the submodule directive. @mbavaro can we revert to the previous method (`https`) so that the diff is only the tag updates + submodule directive + header include?

PR #1284, opened 2019-07-03
Route smoothpose the cov messages not the standard ones

anth-cole:
@benjaminaltieri Trying to pipe the POS_LLH_COV message into smoothpose - can you sanity check this?

@denniszollo FYI


benjaminaltieri:
values look fine @anth-cole , there will need to be a correlating PR in `pose_daemon_wrapper` for it to ingest these. does that exist yet? probably better to get that hash updated so you can run smoothpose scenarios in HITL and make sure things are working properly

anth-cole:
thanks @benjaminaltieri was trying to build locally the pose wrapper first, but I gave up on that

@denniszollo this seems to build now and run - I'll see what hitl has to say but as we spoke yesterday may not be worth it to run

PR #1301, opened 2019-08-14
Dzollo/advanced navigation wheel odo gnss time

denniszollo:


package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -50,29 +51,67 @@
...
+    odo_msg->flags = (u8)(SBP_MSG_ODOMETRY_FLAGS_SRC0 | SBP_MSG_ODOMETRY_FLAGS_SYSTIME);
+  }
 }
 
 static void send_sbp_odometry_msg(odometer_packet_t *an_odo_packet)
 {
   msg_odometry_t odo_msg;
   convert_anpp_odometer_to_sbp(an_odo_packet, &odo_msg);
-  sbp_message_send(SBP_MSG_ODOMETRY, sizeof(odo_msg), (u8 *)&odo_msg, sbp_sender_id_get(), NULL);
+  if (last_gpstime_tow != INVALID_TOW) {
```
benjaminaltieri: Let's put this above `convert_anpp_odometer_to_sbp`, then we can remove the time check which isn't part of the conversion logic

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -50,29 +51,67 @@
 #define SBP_MSG_ODOMETRY_FLAGS_SRC2 (0x08u)
 #define SBP_MSG_ODOMETRY_FLAGS_SRC3 (0x0Cu)
 
-bool anpp_debug = false;
+#define INVALID_TOW 0xffffffff
+#define TIME_SOURCE_MASK 0x07 /* Bits 0-2 */
+#define NO_TIME 0
+#define MS_IN_WEEK 7 * 24 * 60 * 60 * 1000
```
benjaminaltieri: I would suggest wrapping these macro defines with parens and giving a type specifier

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -50,29 +51,67 @@
...
 
+bool anpp_debug = false;
```
benjaminaltieri: Maybe we can wrap these in a context and pass to them to callbacks instead of accessing as a bare global. Also could add debug as a setting in case we want dynamic troubleshooting capabilities

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -50,29 +51,67 @@
...
+    struct timespec now;
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    /* Add on elapsed time since gps_time_callback was received */
+    tow += (now.tv_sec * 1000.0 + now.tv_nsec / 1000000.0)
+           - (last_gps_systime.tv_sec * 1000.0 + last_gps_systime.tv_nsec / 1000000.0);
+    tow += odo_time_offset_ms;
+    if (tow > MS_IN_WEEK) {
+      tow -= MS_IN_WEEK;
+    }
+    if (tow < 0) {
```
benjaminaltieri: Just curious which case this is catching, if tow rolls over between the last GPS time and odo sample? Still not super familiar with tow rollover behavior

denniszollo: This is catching if my addition / subtraction for the time puts is into the previous or the next week.  If the addend I added to the time of week has more than the number of seconds in the week, then I need to wrap around zero.  If it is negative, I need to add to it.

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
denniszollo: yes that is a good idea, will do.

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
denniszollo: willco.

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
denniszollo: This is to ensure that the time of week in our output odo message is between 0 and MS_IN_WEEK.  After I add or subtract from the incoming time of week based upon time of receipt, I could make a negative time of week or one that has more than MS_IN_WEEK.  I think these statements guard against that.

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
denniszollo: agreed.

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -51,30 +51,44 @@
...
-unsigned int last_gpstime_tow = INVALID_TOW;
-struct timespec last_gps_systime;
-int odo_time_offset_ms = 0;
+
+typedef struct anpp_bridge_ctx_s {
+  unsigned int last_gpstime_tow;
+  struct timespec last_gps_systime;
+  int odo_time_offset_ms;
+  pk_endpoint_t *anpp_sub;
+} anpp_bridge_ctx_t;
```
benjaminaltieri: you can use struct initializers here instead of using an explicit init function:

package/sbp_anpp_bridge/src/sbp_anpp_bridge.c
```diff
@@ -83,44 +97,45 @@ static int notify_time_offset_changed(void *context)
...
-    }
-    if (tow < 0) {
-      tow += MS_IN_WEEK;
-    }
-    odo_msg->tow = (u32)tow;
-    odo_msg->velocity = (s32)M_TO_MM(an_odo_packet->speed);
-    odo_msg->flags = (u8)(SBP_MSG_ODOMETRY_FLAGS_SRC0 | SBP_MSG_ODOMETRY_FLAGS_SYSTIME);
+  double tow = context->last_gpstime_tow;
+  struct timespec now;
+  clock_gettime(CLOCK_MONOTONIC, &now);
```
benjaminaltieri: maybe we can foist this helper from `manage_led.c` into `libpiksi/util`
https://github.com/swift-nav/piksi_buildroot/blob/e01649bcf3cf27b69cf67fa3f070ca831bab8bfc/package/piksi_leds/src/manage_led.c#L103

PR #1312, opened 2019-08-30
Forward heading

anth-cole:
@denniszollo @benjaminaltieri - This turned out to be a simple port of ben's original PR, seems to be complete but haven't done any testing


benjaminaltieri:
had to regenerate the configs, this has changed slightly since you last used it I'm guessing @anth-cole

benjaminaltieri:
will also need to update the smoothpose router yaml in case this is targeted at smoothpose operation (I'm assuming not?)

anth-cole:
thanks @benjaminaltieri and nope, don't think any smoothpose routings need to change - it should listen for incoming heading messages and rebroadcast, smoothpose _should_ be independent

benjaminaltieri:
actually what I meant was that the router config that gets activated when smoothpose is enabled would need the same change as the vanilla `sbp_router.yml`

anth-cole:
ahh - gotcha - thanks!

denniszollo:
I am going to do some quick testing on my bench setup to make sure things work in practice.

anth-cole:
I think to test this we need to do the following

- Disable the new heading forwarder setting and check the device operates normally (both time matched and low latency modes, SPP and RTK and with smoothpose enabled and disabled)
- Enable the heading forwarding feature, set the device into RTK mode with a CORS station/Skylark and connect to a second piksi operating in heading mode and check the first piksi rebroadcasts the heading message.
- Repeat above only have both piksi's operating in heading mode and check the first piksi is generating heading and forwarding the second heading message correctly.

PR #1331, opened 2019-11-19
Remove post-tag scripts since they're no longer used [master]

silverjam:
Bring `master` branch of `piksi_buildroot` in line with other repos.  This logic is handled in `piksi-releases` now.


mfine:
If this repo is dead for forward development, this shouldn't be needed, since it'll never be released.

benjaminaltieri (commented):
assuming we should close based on @mfine's comment?

mfine:
You can pull it in or close - shouldn't matter.

PR #1355, opened 2020-02-04
V2.3.0 cn stillness hotfix

denniszollo:


nsirola:
Note, investigating SBAS TTFF degradation with this build in https://swift-nav.atlassian.net/browse/DML-27

